PR #204, opened 2019-07-11
Mark unscannable pages

haonliu:
#### Description of changes

Before we run a11y scan on a page, I'm making the change to check the response header of the page, if it is not 'text/html', we return error information and skip the scan.

Then the page document will have a flag ``unscannable = true`` in its last run property.

#### Pull request checklist

### DIFF
--- a/packages/runner/src/scanner/page.ts
+++ b/packages/runner/src/scanner/page.ts
@@ -61,6 +61,7 @@ export class Page {
     }
 
     private getContentType(headers: Record<string, string>): string {
-        return headers['content-type'] || headers['Content-Type'];
+        // All header names are lower-case, According to puppeteer API doc
+        return headers['content-type'];
     }
 }
--- a/packages/runner/src/scanner/page.ts
+++ b/packages/runner/src/scanner/page.ts
@@ -46,6 +46,7 @@ export class Page {
 
         const axePuppeteer: AxePuppeteer = this.axePuppeteerFactory(this.puppeteerPage);
         const scanResults = await axePuppeteer.analyze();
+
         return { results: scanResults };
     }
 
@@ -57,7 +58,8 @@ export class Page {
 
     private isHtmlPage(response: Puppeteer.Response): boolean {
         const contentType = this.getContentType(response.headers());
-        return contentType != null && contentType.indexOf('text/html') != -1;
+
+        return contentType !== undefined && contentType.indexOf('text/html') !== -1;
     }
 
     private getContentType(headers: Record<string, string>): string {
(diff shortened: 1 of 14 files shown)


packages/runner/src/scanner/page.ts
```diff
@@ -27,28 +28,39 @@ export class Page {
...
         }
     }
+
+    private isHtmlPage(response: Puppeteer.Response): boolean {
+        const contentType = this.getContentType(response.headers());
+        return contentType != null && contentType.indexOf('text/html') != -1;
+    }
+
+    private getContentType(headers: Record<string, string>): string {
+        return headers['content-type'] || headers['Content-Type'];
```
devaradhanm: >headers['content-type'] || headers['Content-Type']; [](start = 15, length = 51)

or something like this:
Object.keys(headers).filter(p => p.toLowercase() == content-type).first();

haonliu: Discussed offline, no need to check the string case here since they mentioned ``All header names are lower-case`` in [puppeteer API doc](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#responseheaders)

packages/runner/src/scanner/scanner.ts
```diff
@@ -16,11 +16,8 @@ export class Scanner {
 
             await this.page.create();
             await this.page.enableBypassCSP();
-            await this.page.goto(url);
 
-            const result = await this.page.scanForA11yIssues();
-
-            return { results: result };
+            return this.page.scanForA11yIssues(url);
```
manishsat: Use return await otherwise catch will never gets executed. Here is more info 
https://jakearchibald.com/2017/await-vs-return-vs-return-await/

haonliu: Thanks, good catch!

packages/runner/src/tasks/data-factory-task.ts
```diff
@@ -33,7 +33,7 @@ export class DataFactoryTask {
 
             return { results: scanResults };
         } else {
-            return { results: [], error: axeScanResults.error };
+            return { results: [], error: axeScanResults.error, unscannable: axeScanResults.unscannable === true ? true : false };
```
manishsat: Is it ok to just use unscannable: axeScanResults.unscannable, whats harm in having scannable as  'undefined'?

haonliu: It is OK I think.

manishsat:
LGTM

PR #207, opened 2019-07-13
adding runtime service configuration

devaradhanm:
This PR has changes to add the service config class in common package.
In the next PR, i will be making changes to consume this file & also copy the file to blob container.


packages/common/src/configuration/runtime-config/runtime-config.dev.json
```diff
@@ -0,0 +1,14 @@
+{
+    "commonConfig": {
+        "taskTimeout": 5,
```
lamaks: >taskTimeout [](start = 9, length = 11)

taskTimeoutInMinutes

packages/common/src/configuration/service-configuration.ts
```diff
@@ -0,0 +1,45 @@
...
+import * as fs from 'fs';
+import { injectable } from 'inversify';
+import { isNil } from 'lodash';
+import * as defaultConfig from './runtime-config/runtime-config.dev.json';
+
+@injectable()
+export class ServiceConfiguration {
+    public static readonly profilePath = './runtime-configuration/runtime-configuration.json';
+    public config: typeof defaultConfig;
+    private readonly fsObj: typeof fs;
```
lamaks: >fsObj [](start = 21, length = 5)

rename to a file?

packages/common/src/configuration/runtime-config/runtime-config.dev.json
```diff
@@ -0,0 +1,14 @@
+{
```
lamaks: Add type of a config: PROD, DEV, PPE, etc.

packages/common/src/configuration/runtime-config/runtime-config.dev.json
```diff
@@ -0,0 +1,14 @@
...
+    "commonConfig": {
```
lamaks: >  [](start = 0, length = 1)

Should we move this file under \packages\config ?

packages/common/src/configuration/service-configuration.ts
```diff
@@ -0,0 +1,45 @@
...
+                        this.fsObj.readFile(ServiceConfiguration.profilePath, (err, buffer) => {
+                            if (isNil(err)) {
+                                this.config = JSON.parse(buffer.toString('utf-8')) as typeof defaultConfig;
+                                resolve();
+                            } else {
+                                reject(err);
+                            }
+                        });
+                    } else {
+                        console.log('Unable to find runtime configuration file. Using dev configuration');
```
lamaks: >console.log( [](start = 24, length = 12)

Can echo to console the config file contents

lamaks: >Using dev configuration [](start = 80, length = 23)

We may need to require mandatory env config ie fail if prod config not available. Can also intro the process var with env type (prod, dev, ppe, etc.)

devaradhanm: checking for dev.json file instead instead of embedding the file in webpack.

packages/common/src/configuration/runtime-config/runtime-config.dev.json
devaradhanm: i didnt want to create a package for every small feature that we have. we can move this later to config package based on complexity.

packages/common/src/configuration/service-configuration.ts
devaradhanm: renamed to fileSystem

packages/common/src/configuration/service-configuration.ts
```diff
@@ -0,0 +1,110 @@
...
+        key?: K,
+    ): Promise<K extends null | undefined ? RuntimeConfig : RuntimeConfig[K]> {
+        const config = await this.getConvictConfig();
+
+        return config.get(key);
+    }
+
+    private async getConvictConfig(): Promise<convict.Config<RuntimeConfig>> {
+        if (isNil(this.loadConfigPromise)) {
+            this.loadConfigPromise = new Promise((resolve, reject) => {
```
lamaks: >Promise((resolve, reject) [](start = 41, length = 25)

Do we need to also reject on exception

devaradhanm: if the callback of promise throws then the promise will be rejected automatically. i have added a test to cover this now.

PR #210, opened 2019-07-17
using max queue size config from runtime config

devaradhanm:
#### Description of changes

As part of this, I have also refactored dispatcher tests to improve tests to cover all cases

#### Pull request checklist

### DIFF
--- a/packages/scan-request-sender/src/sender/dispatcher.ts
+++ b/packages/scan-request-sender/src/sender/dispatcher.ts
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { client, CosmosOperationResponse } from 'azure-services';
+import { ServiceConfiguration } from 'common';
 import { inject, injectable } from 'inversify';
 import { Logger } from 'logger';
 import { PageDocumentProvider } from 'service-library';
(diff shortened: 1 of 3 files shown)


packages/scan-request-sender/src/sender/dispatcher.ts
```diff
@@ -13,10 +14,11 @@ export class Dispatcher {
         @inject(PageDocumentProvider) private readonly pageDocumentProvider: PageDocumentProvider,
         @inject(Logger) private readonly logger: Logger,
         @inject(ScanRequestSender) private readonly sender: ScanRequestSender,
+        @inject(ServiceConfiguration) private readonly serviceConfig: ServiceConfiguration,
     ) {}
 
     public async dispatchScanRequests(): Promise<void> {
-        const configQueueSize = Number(process.env.QUEUE_SIZE);
+        const configQueueSize = (await this.serviceConfig.getConfigValue('commonConfig')).maxQueueSize;
```
manishsat: the api call should be serviceConfig.getMaxQueueSize() or something similar.
getConfigValue('commonConfig') --> this shows that user of the class should aware of the section of the config.

devaradhanm: i didnt want to create a method for every property that exists in the config. So, user can get a property under a group object like scanConfig / commonConfig etc.

lamaks: >'commonConfig' [](start = 73, length = 14)

Can we use generic instead of string?
serviceConfig.getConfigValue<configType>().value

devaradhanm: even though we are using string, it is type safe. So, if you call serviceConfig.getConfigValue('blah'), it will fail

devaradhanm: also, i didnt want to add a method for every property in the service config.

lamaks: how we discover the name then?

---
In reply to: [304637452](https://github.com/microsoft/accessibility-insights-service/pull/210#discussion_r304637452) [](ancestors = 304637452)

PR #221, opened 2019-07-20
Added pool tasks processing evaluation formula

lamaks:
#### Description of changes

Added pool tasks processing evaluation formula

#### Pull request checklist

### DIFF
--- a/packages/job-manager/src/batch/batch.ts
+++ b/packages/job-manager/src/batch/batch.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 import { BatchServiceModels } from '@azure/batch';
 import { Message } from 'azure-services';
-import { ServiceConfiguration } from 'common';
+import { JobManagerConfig, ServiceConfiguration } from 'common';
 import * as crypto from 'crypto';
 import { inject, injectable } from 'inversify';
 import * as _ from 'lodash';
@@ -13,6 +13,8 @@ import { BatchServiceClientProvider, jobManagerIocTypeNames } from '../job-manag
 import { BatchConfig } from './batch-config';
 import { JobTask, JobTaskState } from './job-task';
 import { RunnerTaskConfig } from './runner-task-config';
+
+// tslint:disable: max-line-length
 
 @injectable()
 export class Batch {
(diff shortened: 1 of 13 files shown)


packages/job-manager/src/batch/batch.ts
```diff
@@ -26,6 +28,61 @@ export class Batch {
...
+        // The script will calculate the tasks processing ratio returned in $taskProcessingRatio variable as
+        // double value and will return -1 when there are no active tasks or sampling data is insufficient
+
+        // the time interval for data sampling
+        $sampleTimeIntervalMinutes = 15 * TimeInterval_Minute;
+        $samplePercentThreshold = 70;
+
+        // get sample data for pending tasks (the sum of active and running tasks)
+        $pendingTasksSamplePercent = $PendingTasks.GetSamplePercent($sampleTimeIntervalMinutes);
+        $pendingTasksVector = $PendingTasks.GetSample(TimeInterval_Minute, $sampleTimeIntervalMinutes);
```
devaradhanm: the documentation says, GetSample may fail. does this mean, it may throw? if so, do we need to handle that? 

(under https://docs.microsoft.com/en-us/azure/batch/batch-automatic-scaling#getsampledata)

Because the GetSample method fails if the percentage of samples returned is less than the samplePercent specified, you can use the GetSamplePercent method to check first. Then you can perform an alternate action if insufficient samples are present, without halting the automatic scaling evaluation.

packages/job-manager/src/batch/batch.ts
```diff
@@ -26,6 +28,61 @@ export class Batch {
...
+        $pendingTasksAvg = avg($pendingTasksVector);
+
+        // get sample data for running tasks
+        $runningTasksSamplePercent = $RunningTasks.GetSamplePercent($sampleTimeIntervalMinutes);
+        $runningTasksVector = $RunningTasks.GetSample(TimeInterval_Minute, $sampleTimeIntervalMinutes);
+        $runningTasksAvg = avg($runningTasksVector);
+
+        // calculate the tasks processing ratio
+        $taskProcessingRatio = $runningTasksAvg / ($pendingTasksAvg > 0 ? $pendingTasksAvg : 1);
```
devaradhanm: should this be?
```suggestion
        $taskProcessingRatio = ($runningTasksAvg > 0 || $pendingTasksAvg > 0) ? $runningTasksAvg / ($pendingTasksAvg +  $runningTasksAvg) : 0;
```

packages/job-manager/src/batch/batch.ts
```diff
@@ -26,6 +28,61 @@ export class Batch {
...
     ) {}
 
+    public async getPoolTaskProcessingRatio(): Promise<number> {
+        const resultVariableName = 'taskProcessingRatioResult';
+        const poolTaskProcessingRatioFormula = `
+        // The script will calculate the tasks processing ratio returned in $taskProcessingRatio variable as
+        // double value and will return -1 when there are no active tasks or sampling data is insufficient
+
+        // the time interval for data sampling
+        $sampleTimeIntervalMinutes = 15 * TimeInterval_Minute;
```
devaradhanm: is this formula something we will be changing at runtime? if not shouldnt this be part of arm template / deployment?
also, autoscaling will break if you go beyond 50 nodes / downscale to 0 & then scale up.

lamaks: We will not change this at runtime and this does not belong to the deployment. Also it has no connection with 50's node limit issue. It uses batch stat data to calculate average pool load. Note this is not auto scale implementation.

---
In reply to: [305949078](https://github.com/microsoft/accessibility-insights-service/pull/221#discussion_r305949078) [](ancestors = 305949078)

packages/job-manager/src/batch/batch.ts
lamaks: The formula evaluation will not throw. The documentation you referenced explains case where samplePercent parameter is specified. The current code does not have that parameter.

---
In reply to: [305942247](https://github.com/microsoft/accessibility-insights-service/pull/221#discussion_r305942247) [](ancestors = 305942247)

packages/job-manager/src/batch/batch.ts
lamaks: taskProcessingRatio is percent ratio of running tasks to all tasks in progress (that is sum of active and running tasks). This line make sure that there is no division on 0. Next line will return  -1 when no tasks been processed ie pendingTaskCount = 0

---
In reply to: [305944851](https://github.com/microsoft/accessibility-insights-service/pull/221#discussion_r305944851) [](ancestors = 305944851)

packages/job-manager/src/batch/batch.ts
```diff
@@ -26,6 +28,61 @@ export class Batch {
...
+        $samplePercentThreshold = 70;
+
+        // get sample data for pending tasks (the sum of active and running tasks)
```
devaradhanm: why not call batch apis to know number of idle nodes & number of running nodes instead of using the auto scale formula?

PR #223, opened 2019-07-30
adding runtime config package that updates config on blob update

devaradhanm:
#### Description of changes

I still havn't removed the serviceConfiguration class from common package & its usages.
The new package runtime-config-provider provides serviceConfig class that listens to blob changes & returns the latest config.

I will send another PR to start using this package.

For review, Core changes are in: BlobReader, ConfigFileUpdater. ServiceConfiguration class is almost same as the class in common package, except that it listens to ConfigFileUpdater.

#### Pull request checklist

### DIFF
--- a/packages/common/src/timers/async-interval.ts
+++ b/packages/common/src/timers/async-interval.ts
@@ -1,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { injectable } from 'inversify';
+
+@injectable()
+export class AsyncInterval {
+    /**
+     * Executes the given callback after the given interval time in loop,
+     * till the callback returns value other than true
+     */
+    public setIntervalExecution(callback: () => Promise<boolean>, intervalInMilliSec: number): void {
+        setTimeout(async () => {
+            try {
+                const response = await callback();
+                if (response === true) {
+                    this.setIntervalExecution(callback, intervalInMilliSec);
+                }
+            } catch (e) {
+                console.log('error occured in interval execution - ', e);
+                this.setIntervalExecution(callback, intervalInMilliSec);
+            }
+        }, intervalInMilliSec);
+    }
+}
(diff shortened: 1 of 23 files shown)


packages/common/src/timers/async-interval.ts
```diff
@@ -0,0 +1,25 @@
...
+    public setIntervalExecution(callback: () => Promise<boolean>, intervalInMilliSec: number): void {
+        setTimeout(async () => {
+            try {
+                const response = await callback();
+                if (response === true) {
+                    this.setIntervalExecution(callback, intervalInMilliSec);
+                }
+            } catch (e) {
+                console.log('error occured in interval execution - ', e);
+                this.setIntervalExecution(callback, intervalInMilliSec);
```
manishsat: Any reason why you called setIntervalExecution again for the callback if errored out? setTimeout will do it automatically.

packages/runtime-config-provider/src/blob-reader.ts
```diff
@@ -0,0 +1,79 @@
...
+    notFound: boolean;
+    updatedContent: string;
+}
+
+@injectable()
+export class BlobReader {
+    private readonly blobServiceClient: BlobServiceClient;
+
+    constructor(@inject(runtimeConfigIocTypes.BlobServiceClientFactory) blobServiceClientProvider: BlobServiceClientFactory) {
+        this.blobServiceClient = blobServiceClientProvider();
```
manishsat: Would be good if you do something like blobServiceClientProvider.createBlogServiceClient(), it is more readable in my opinion.

packages/common/src/timers/async-interval.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { injectable } from 'inversify';
+
+@injectable()
+export class AsyncInterval {
+    /**
+     * Executes the given callback after the given interval time in loop,
+     * till the callback returns value other than true
+     */
+    public setIntervalExecution(callback: () => Promise<boolean>, intervalInMilliSec: number): void {
```
lamaks: >intervalInMilliSec [](start = 66, length = 18)

Milliseconds

PR #227, opened 2019-07-31
Add script to upload only config files

AhmedAbdoOrtiga:
#### Description of changes
- Move config files upload in a separate script.
- Call that script in upload files script.
- Exclude .snap files from prettier.

#### Pull request checklist

### DIFF
--- a/packages/resource-deployment/scripts/upload-files.sh
+++ b/packages/resource-deployment/scripts/upload-files.sh
@@ -9,8 +9,6 @@ export jobManagerContainerName="batch-job-manager-script"
 export runnerContainerName="batch-runner-script"
 export scanRequestSenderContainerName="batch-scan-request-sender-script"
 export poolStartupContainerName="batch-pool-startup-script"
-export runtimeConfigurationContainerName="runtime-configuration"
-export runtimeConfigurationBlobName="runtime-config.json"
 export includePattern="*[!*.map]"
 
 if [[ -z $dropFolder ]]; then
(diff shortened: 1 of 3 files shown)


packages/resource-deployment/scripts/upload-files.sh
```diff
@@ -68,4 +57,4 @@ uploadFolderContents $jobManagerContainerName "$dropFolder/job-manager/dist" "$s
 uploadFolderContents $runnerContainerName "$dropFolder/runner/dist" "$storageAccountName" "$includePattern"
 uploadFolderContents $scanRequestSenderContainerName "$dropFolder/scan-request-sender/dist" "$storageAccountName" "$includePattern"
 uploadFolderContents $poolStartupContainerName "$dropFolder/resource-deployment/dist/scripts/pool-startup" "$storageAccountName" "$includePattern"
-uploadFile $runtimeConfigurationContainerName "$dropFolder/resource-deployment/dist/runtime-config/runtime-config.$environment.json" "$storageAccountName" "$runtimeConfigurationBlobName"
+. "${0%/*}/upload-config-files.sh"
```
flyingUnderTheRadar: Will this script need to be passed some parameters or is there another way that it gets these parameters (env variables etc)?

AhmedAbdoOrtiga: when you call a script inside another using the same way . "filepath"
the second script gets executed inside the first script process, all function and variables in the first script are accessible in the second, so it will use the same input parameters for the first script.

PR #232, opened 2019-08-09
Scan REST API contract

lamaks:
#### Description of changes

Swagger mock https://msswagger/designer/apis/sce4ik/publishedmock

#### Pull request checklist

### DIFF
--- a/packages/api-contracts/ai-scan-api.yaml
+++ b/packages/api-contracts/ai-scan-api.yaml
@@ -1,0 +1,232 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+openapi: 3.0.0
+info:
+    version: 1.0.0
+    title: Accessibility Insight Service – Scan API
+paths:
+    /scans:
+        post:
+            summary: Create accessibility scan requests
+            operationId: createScans
+            parameters:
+                - name: api-version
+                  in: path
+                  required: true
+                  description: The client API version
+                  schema:
+                      type: string
+                  example: '1.0'
+            requestBody:
+                content:
+                    application/json:
+                        schema:
+                            $ref: '#/components/schemas/ScanRunRequests'
+            responses:
+                '201':
+                    description: Created
+                    content:
+                        application/json:
+                            schema:
+                                $ref: '#/components/schemas/ScanRunResponses'
+    /scans/{scanid}:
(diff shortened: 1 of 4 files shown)


packages/api-contracts/ai-scan-api.yaml
```diff
@@ -0,0 +1,232 @@
...
+                      type: string
+                  example: '1.0'
+            responses:
+                '200':
+                    description: OK
+                    content:
+                        application/json:
+                            schema:
+                                $ref: '#/components/schemas/ScanResultResponse'
+    /scans/$batch:
```
devaradhanm: what is the mock value for batch?

lamaks: no value this is uri path

packages/api-contracts/ai-scan-api.yaml
```diff
@@ -0,0 +1,232 @@
...
+                  schema:
+                      type: string
+                  example: '1.0'
+            requestBody:
+                content:
+                    application/json:
+                        schema:
+                            $ref: '#/components/schemas/ScanRunRequests'
+            responses:
+                '201':
```
flyingUnderTheRadar: Should we consider using HTTP 202 instead of HTTP 201?
202 indicates that the request was accepted but a resource was not created. [More Info](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202)

packages/api-contracts/ai-scan-api.yaml
```diff
@@ -0,0 +1,232 @@
...
+                  description: The client API version
+                  schema:
+                      type: string
+                  example: '1.0'
+            requestBody:
+                content:
+                    application/json:
+                        schema:
+                            $ref: '#/components/schemas/ScanRuns'
+            responses:
```
flyingUnderTheRadar: Should we also add error responses, maybe 404 / 410 here so that error contracts are also defined?

Here and elsewhere as well if needed.

lamaks: We are going to use plain HTTP 404 response. No need to define special contract for now.

packages/api-contracts/ai-scan-api.yaml
lamaks: Updated to 202

PR #233, opened 2019-08-16
Scan REST API with mock responses

lamaks:
#### Description of changes

Scan REST API with mock responses. The PR implements the AI service scan REST API defined in https://github.com/microsoft/accessibility-insights-service/tree/master/packages/api-contracts yaml spec bu creating Azure Functions endpoints for each REST API.

#### Pull request checklist


packages/common/src/system/nullable-value.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export interface NullableValue<T> {
```
devaradhanm: why do we need this C# like type in javascript, when js can support null & non null value for any object?

packages/azure-services/src/azure-cosmos/cosmos-client-wrapper.ts
```diff
@@ -157,7 +157,7 @@ export class CosmosClientWrapper {
         try {
             const options: cosmos.RequestOptions = this.getRequestOptionsWithPartitionKey(partitionKey);
             const response = await container.item(id).read(options);
-            const itemT = this.convert<T>(response.body);
+            const itemT = System.convert<T>(response.body);
```
devaradhanm: >System.convert<T>(response.body); [](start = 26, length = 33)

why is this call needed, when all it does is just cast it to type T?

devaradhanm:
>     public convert<T>(source: any, typeT?: new () => T): T {

why do you need to have a case where typeT constructor is not passed? since u are just type casting in that case?

---
Refers to: packages/common/src/system/activator.ts:12 in 6d9f27f. [](commit_id = 6d9f27ffc5098a6990094db746836a221e1c2110, deletion_comment = False)

packages/job-manager/tsconfig.json
```diff
@@ -1,6 +1,10 @@
 {
     "extends": "../../tsconfig.base.json",
+    "lib": ["es6", "dom"],
+    "types": ["reflect-metadata"],
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
     "compilerOptions": {
-        "outDir": "out"
+        "outDir": "dist"
     }
```
devaradhanm: will this override all compilerOptions set on tsconfig.base? (I am assuming it won't)
why not move this to base, so that, you don't have to repeat this everywhere?

packages/logger/tsconfig.json
```diff
@@ -1,5 +1,9 @@
 {
     "extends": "../../tsconfig.base.json",
+    "lib": ["es6", "dom"],
+    "types": ["reflect-metadata"],
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
```
devaradhanm: shouldnt this be in base?

packages/web-api/src/controllers/post-scans-controller.ts
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { System } from 'common';
+import { ScanRunRequest } from '../api-contracts/scan-run-request';
+import { ScanRunResponse } from '../api-contracts/scan-run-response';
+import { ApiController } from './api-controller';
+
+export class PostScansController extends ApiController {
```
devaradhanm: >PostScansController [](start = 13, length = 19)

may be it might be better to have ScannController(to submit & receive scan info) & ReportController (to download report)

packages/web-api/src/controllers/api-controller.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+    public readonly apiVersion = '1.0';
+    public invoked = false;
+
+    protected invokeImpl(): void {
+        this.invoked = true;
+    }
+}
+
+describe('validateContentType()', () => {
+    it('should validate HTTP POST and PUT only', () => {
```
devaradhanm: > it('should validate HTTP POST and PUT only', () => { [](start = 3, length = 53)

you should also have a test for PUT request

packages/web-api/src/controllers/api-controller.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+                method: 'POST',
+                rawBody: `{ id: '1' }`,
+                headers: { host: 'localhost' },
+            },
+        });
+        const apiControllerMock = new ApiControllerMock(context);
+        const valid = apiControllerMock.validateContentType();
+        expect(context.res).toEqual({ status: 400, body: 'Content type was not specified' });
+        expect(valid).toEqual(false);
+    });
```
devaradhanm: you could consider using data driven test

packages/common/src/system/nullable-value.ts
lamaks: Removed

packages/web-api/src/controllers/post-scans-controller.ts
lamaks: Merge to single scan controller

packages/web-api/src/controllers/api-controller.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+                method: 'GET',
+            },
+        });
+        const apiControllerMock = new ApiControllerMock(context);
+        const valid = apiControllerMock.validateContentType();
+        expect(context.res).toEqual(undefined);
+        expect(valid).toEqual(true);
+    });
+
+    it('should validate empty body', () => {
```
devaradhanm: >validate [](start = 15, length = 8)

instead of naming the function as validate, mention the expected outcome. foreg: should fail for empty body.
or shopuld pass for valid body content etc.

packages/web-api/src/controllers/get-report-controller.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { getSarifReportMock } from '../../get-report-func/sarif-report-mock';
```
devaradhanm: controllers should not interact with func folder. but it is fine if you don't want to fix this since we will be removing in the next feature

packages/job-manager/tsconfig.json
lamaks: moving to base breaks repo build. need more investigation.

packages/logger/tsconfig.json
lamaks: moving to base breaks repo build. need more investigation.

packages/common/src/system/activator.spec.ts
```diff
@@ -24,6 +26,15 @@ describe('Activator', () => {
         expect(instance).toEqual('value');
     });
 
+    it('convert to array type', () => {
+        const source: any[] = [{ p1: 'value' }];
+
+        const instance = activator.convert<string>(source);
```
haonliu: Should this be ``string[]``?

packages/web-api/src/controllers/api-controller.spec.ts
lamaks: updated

packages/web-api/src/controllers/api-controller.spec.ts
lamaks: updated

packages/common/src/system/activator.spec.ts
lamaks: updated

packages/web-api/src/controllers/get-report-controller.ts
lamaks: updated

packages/azure-services/src/azure-cosmos/cosmos-client-wrapper.ts
lamaks: remove activator

PR #241, opened 2019-08-23
[WIP] Create Script for creating Function App

AhmedAbdoOrtiga:
#### Description of changes
- Create a script for creating a function app using ARM template

TODO:
- Deploy the web-api functions to the created function app.

#### Pull request checklist

### DIFF
--- a/packages/resource-deployment/scripts/function-app-create.sh
+++ b/packages/resource-deployment/scripts/function-app-create.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+set -euo pipefail
+
+exitWithUsageInfo() {
+    echo "
+Usage: $0 -s <subscription name or id> -r <resource group> -l <location>
+"
+    exit 1
+}
+
+export subscription=""
+export resourceGroupName=""
+export location=""
+export functionAppName
+export appInsightsName
+export storageAccountName
+
+# Set default ARM Function App template files
+templateFilePath="../templates/function-app-template.json"
+
+# Read script arguments
+while getopts "s:r:l:"  option; do
+    case $option in
+		s) subscription=${OPTARG} ;;
+		r) resourceGroupName=${OPTARG} ;;
+		l) location=${OPTARG} ;;
+		*) exitWithUsageInfo ;;
+		esac
+done
+
+if [ -z "$subscription" ] || [ -z "$resourceGroupName" ] || [ -z "$location" ]; then
+	echo "Either one of subscription, resourceGroupName or location is empty"
+	exitWithUsageInfo
+fi
+
+# Login to Azure account if required
+if ! az account show 1>/dev/null; then
+    az login
+fi
+
+#set the default subscription id
+az account set --subscription $subscription
+
+#Check for existing RG
+if [ $(az group exists --name $resourceGroupName) = false ]; then
+	echo "Resource group with name" $resourceGroupName "could not be found. Creating new resource group.."
+	az group create --name $resourceGroupName --location $location 1> /dev/null
+	echo "Resource group has been successfully created"
+else
+	echo "Using existing resource group..."
+fi
+
+#Start deployment
+echo "Deploying Function App using ARM template"
+export resourceName
+resources=$(az group deployment create \
+    --resource-group "$resourceGroupName" \
+    --template-file "$templateFilePath" \
+    --query "properties.outputResources[].id" \
+    -o tsv)
+
+
+. "get-resource-name-from-resource-paths.sh" -p "Microsoft.Web/sites" -r "$resources"
+functionAppName=$resourceName
+echo "Successfully deployed Function App '$functionAppName'"
+
+. "get-resource-name-from-resource-paths.sh" -p "Microsoft.Storage/storageAccounts" -r "$resources"
+storageAccountName=$resourceName
+echo "Successfully created Storage Account '$storageAccountName'"
+
+. "get-resource-name-from-resource-paths.sh" -p "microsoft.insights/components" -r "$resources"
+appInsightsName=$resourceName
+echo "Successfully created Application Insights '$appInsightsName'"
--- a/packages/resource-deployment/templates/function-app-template.json
+++ b/packages/resource-deployment/templates/function-app-template.json
@@ -1,0 +1,103 @@
+{
+    "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "name": {
+            "type": "string",
+            "defaultValue": "a11yfunctionapp",
+            "metadata": {
+                "description": "The location in which the Azure Storage resources should be deployed."
+            }
+        },
+        "storageAccountName": {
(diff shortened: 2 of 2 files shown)


packages/resource-deployment/templates/function-app-template.json
```diff
@@ -0,0 +1,103 @@
+{
+    "$schema": "http://schema.management.azure.com/schemas/2014-04-01-preview/deploymentTemplate.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "name": {
+            "type": "string",
+            "defaultValue": "a11yfunctionapp",
+            "metadata": {
+                "description": "The location in which the Azure Storage resources should be deployed."
```
flyingUnderTheRadar: Does the description here need to change?

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -0,0 +1,73 @@
...
+echo "Deploying Function App using ARM template"
+export resourceName
+resources=$(az group deployment create \
+    --resource-group "$resourceGroupName" \
+    --template-file "$templateFilePath" \
+    --query "properties.outputResources[].id" \
+    -o tsv)
+
+
+. "get-resource-name-from-resource-paths.sh" -p "Microsoft.Web/sites" -r "$resources"
```
flyingUnderTheRadar: If I understand correctly we calling this three times right? Is there a way we can store the response and parse it thrice. (Yes I am aware its idempotent, but itd be great not to call it thrice). If we aren't calling this thrice ignore this completely.

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
+set -euo pipefail
```
devaradhanm: you would also need to make install.sh call this

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -0,0 +1,73 @@
...
+Usage: $0 -s <subscription name or id> -r <resource group> -l <location>
+"
+    exit 1
+}
+
+export subscription=""
+export resourceGroupName=""
+export location=""
+export functionAppName
+export appInsightsName
```
devaradhanm: consider using the same app insights account used for batch

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -0,0 +1,73 @@
...
+while getopts "s:r:l:"  option; do
+    case $option in
+		s) subscription=${OPTARG} ;;
+		r) resourceGroupName=${OPTARG} ;;
+		l) location=${OPTARG} ;;
+		*) exitWithUsageInfo ;;
+		esac
+done
+
+if [ -z "$subscription" ] || [ -z "$resourceGroupName" ] || [ -z "$location" ]; then
```
devaradhanm: expect appInsightsName to be set as well

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Login to Azure account if required
+if ! az account show 1>/dev/null; then
+    az login
+fi
+
+#set the default subscription id
+az account set --subscription $subscription
+
+#Check for existing RG
```
devaradhanm: remove this. let this script expect that resource group exists. use the same rg as we use for batch in install.sh

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -0,0 +1,73 @@
...
+    --query "properties.outputResources[].id" \
+    -o tsv)
+
+
+. "get-resource-name-from-resource-paths.sh" -p "Microsoft.Web/sites" -r "$resources"
+functionAppName=$resourceName
+echo "Successfully deployed Function App '$functionAppName'"
+
+. "get-resource-name-from-resource-paths.sh" -p "Microsoft.Storage/storageAccounts" -r "$resources"
+storageAccountName=$resourceName
```
devaradhanm: can we use the same storage account as we use for batch?

packages/resource-deployment/scripts/function-app-create.sh
AhmedAbdoOrtiga: That'd would need a refactor for get-resource-name-from-resource-paths.sh, we can do it in a different PR

AhmedAbdoOrtiga: Actually resolved since we don't need to create resources anymore, we're gonna use the previously created resources

packages/resource-deployment/scripts/function-app-create.sh
AhmedAbdoOrtiga: We can't do that, the only way to do it is the same as we do app-insights-create.sh, we set it in the template file.
so we're gonna use the same name as in app-insights-create.template.json and we are **not** going to create a new app insights resource

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -0,0 +1,71 @@
...
+# Read script arguments
+while getopts "r:"  option; do
+    case $option in
+		r) resourceGroupName=${OPTARG} ;;
+		*) exitWithUsageInfo ;;
+		esac
+done
+
+if [ -z "$resourceGroupName" ]; then
+	echo "Either one of subscription, resourceGroupName or location is empty"
```
lamaks: misleading message, only rg is required as parameter. Either add to options rest of params or update the message

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
+
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+# shellcheck disable=SC1090
+set -eo pipefail
+
+export resourceGroupName=""
```
lamaks: why we need empty value here?

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -0,0 +1,71 @@
...
+# Install the Core Tools package
+sudo apt-get install azure-functions-core-tools
+
+# change directory to the functions folder to publish
+cd "${0%/*}/../../web-api/dist"
+
+#publish the functions to the functionAppName
+func azure functionapp publish $functionAppName --javascript
+
+echo "Successfully published"
```
lamaks: Successfully published functions to the Function App '$functionAppName'

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -0,0 +1,71 @@
...
+sudo apt-get update
+
+# Install the Core Tools package
+sudo apt-get install azure-functions-core-tools
+
+# change directory to the functions folder to publish
+cd "${0%/*}/../../web-api/dist"
+
+#publish the functions to the functionAppName
+func azure functionapp publish $functionAppName --javascript
```
lamaks: can be --node

PR #245, opened 2019-08-23
adding packages for webapi scan request sender, scan job manager & sc…

devaradhanm:
…an runner

#### Description of changes

#### Pull request checklist


.vscode/workspace.code-workspace
```diff
@@ -16,6 +16,18 @@
             "name": "scan-request-sender",
             "path": "../packages/scan-request-sender"
         },
+        {
+            "name": "web-api-scan-request-sender",
```
lamaks: Can we have some other name? web-api is not related I think. Maybe wcp, on-demand, continues, stream, batch or something more related?

devaradhanm: i ddidnt want to use the word wcp (the consumer), as it could change. instead name it that is being used for requests submitted via web api

packages/web-api/package.json
```diff
@@ -3,7 +3,7 @@
     "version": "1.0.0",
     "description": "This project welcomes contributions and suggestions.  Most contributions require you to agree to a Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us the rights to use your contribution. For details, visit https://cla.microsoft.com.",
     "scripts": {
-        "build": "webpack --config ./webpack.config.js --config-name web-api --colors --progress",
```
devaradhanm: adding progress messes up the build log. so, removing it.

.vscode/workspace.code-workspace
lamaks: that can be used not only by web api and it is going to be main workflow after moving to full wcp support. web-api is less appropriate here. Call main- at least

PR #263, opened 2019-08-29
Add Authentication/Authorization for function app 

AhmedAbdoOrtiga:
#### Description of changes

- Add Authentication/Authorization for function app.
- Change managed identity creation to be in the ARM template.
- set web-api functions authlevel from the default "function to "anonymous" so no code would be generated in the the API link.

**Note:**
- Will need to create an App Registration in all our environments to be used in Auth (later we can add a script for automating this process). 
- Any CI/CD pipeline that uses the install script will need to add the new 2 parameters (clientId, tenantId).

#### Pull request checklist

### DIFF
--- a/packages/resource-deployment/scripts/function-app-create.sh
+++ b/packages/resource-deployment/scripts/function-app-create.sh
@@ -21,6 +21,7 @@ addReplyUrlIfNotExist() {
     functionAppName=$2
 
     # Get existing reply urls of the app registration
+    echo "Fetching existing replyUrls..."
     replyUrls=$(az ad app show --id $clientId --query "replyUrls" -o tsv)
     replyUrl="https://${functionAppName}.azurewebsites.net/.auth/login/aad/callback"
 
@@ -31,25 +32,19 @@ addReplyUrlIfNotExist() {
         fi
     done
 
-    echo "Adding replyUrl to app registration..."
+    echo "Adding replyUrl $replyUrl to app registration..."
     az ad app update --id $clientId --add replyUrls $replyUrl
     echo "  Successfully added replyUrl."
 }
 
-createAppRegistrationIfNotExist() {
-    clientId=$1
-    appRegistrationName=$2
-    resourceGroupName=$3
-    environment=$4
+createAppRegistration() {
+    resourceGroupName=$1
+    environment=$2
 
-    if [ ! -z "$clientId" ] && (az ad app show --id "$clientId" 1>/dev/null); then
-        appRegistrationName=$(az ad app show --id "$clientId" --query "displayName" -o tsv)
-        echo "'$appRegistrationName' App Registration with Client ID '$clientId' already exists"
-    else
-        appRegistrationName="allyappregistration-${resourceGroupName}-${environment}"
-        clientId=$(az ad app create --display-name "$appRegistrationName" --query "appId" -o tsv)
-        echo "Successfully created '$appRegistrationName' App Registration with Client ID '$clientId'"
-    fi
+    appRegistrationName="allyappregistration-$resourceGroupName-$environment"
+    echo "Creating a new AppRegistration with display name $appRegistrationName..."
+    clientId=$(az ad app create --display-name "$appRegistrationName" --query "appId" -o tsv)
+    echo "  Successfully created '$appRegistrationName' App Registration with Client ID '$clientId'"
 }
 
 # Set default ARM Function App template files
@@ -69,15 +64,16 @@ if [ -z $resourceGroupName ] || [ -z $environment ]; then
 fi
 
 # Will return the function name if we deployed before the function app template on this resource group
-echo "Checking if the function app already exists..."
+echo "Checking if the function app and AppRegistration already exists..."
 functionAppName=$(az group deployment show -g "$resourceGroupName" -n "function-app-template" --query "properties.parameters.name.value" -o tsv)
+clientId=$(az webapp auth show -n "$functionAppName" -g "$resourceGroupName" --query "clientId" -o tsv) || true
 
-# Check if the Function App name is valid and the Function App exists
-if [ ! -z "$functionAppName" ] && (az functionapp show -n "$functionAppName" -g "$resourceGroupName" 1>/dev/null); then
-    echo "function app does not exist, creating a new one..."
-    # Get the Client (App) ID for the App Registration that's used for this Function App authentication
-    clientId=$(az webapp auth show -n "$functionAppName" -g "$resourceGroupName" --query "clientId" -o tsv)
-    createAppRegistrationIfNotExist $clientId $appRegistrationName $resourceGroupName $environment
+if [[ ! -n $clientId ]]; then
+    createAppRegistration $resourceGroupName $environment
+else
+    echo "AppRegistration already exists. Fetching displayName of the AppRegistration..."
+    appRegistrationName=$(az ad app show --id "$clientId" --query "displayName" -o tsv)
+    echo "  Successfully got AppRegistration display name: $appRegistrationName"
 fi
 
 # Start function app deployment
--- a/packages/resource-deployment/scripts/function-app-create.sh
+++ b/packages/resource-deployment/scripts/function-app-create.sh
@@ -8,13 +8,12 @@ set -eo pipefail
 
 export resourceGroupName
 export clientId
-export tenantId
 export functionAppName
 export resourceName
 
 exitWithUsageInfo() {
     echo "
-Usage: $0 -r <resource group> -c <client id> -t <tenant id>
+Usage: $0 -r <resource group> -c <client id>
 "
     exit 1
 }
@@ -23,11 +22,10 @@ Usage: $0 -r <resource group> -c <client id> -t <tenant id>
 templateFilePath="${0%/*}/../templates/function-app-template.json"
 
 # Read script arguments
-while getopts "r:c:t:" option; do
+while getopts "r:c:" option; do
     case $option in
     r) resourceGroupName=${OPTARG} ;;
     c) clientId=${OPTARG} ;;
-    t) tenantId=${OPTARG} ;;
     *) exitWithUsageInfo ;;
     esac
 done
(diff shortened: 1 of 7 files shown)


packages/resource-deployment/scripts/install.sh
```diff
@@ -21,7 +21,7 @@ export environment
 
 exitWithUsageInfo() {
     echo "
-Usage: $0 -e <environment> -l <Azure region> -o <organisation name> -p <publisher email> -r <resource group> -s <subscription name or id> 
+Usage: $0 -e <environment> -l <Azure region> -o <organisation name> -p <publisher email> -r <resource group> -s <subscription name or id> -c <client id> -t <tenant id>
```
devaradhanm: can you also add description for the 2 parameters u have added

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -32,23 +34,30 @@ if [ -z "$resourceGroupName" ]; then
     exitWithUsageInfo
 fi
 
+appRegistrationName="allyappregistration"
+
+# create the app registration if it doesn't exist
+if ! az ad app show --id $clientId 1>/dev/null; then
```
devaradhanm: either use create --id to pass the client id or use display name to check if the id exists using the az command az ad app list --display-name "<principal_name>"

devaradhanm: i prefer display name option

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -32,23 +34,30 @@ if [ -z "$resourceGroupName" ]; then
...
 functionAppName=$resourceName
 echo "Successfully deployed Function App '$functionAppName'"
 
-# Add system-assigned managed identity to function app
-
-echo "Adding managed identity to Function App '$functionAppName'"
-principalId=$(az webapp identity assign --name "$functionAppName" --resource-group "$resourceGroupName" --query principalId -o tsv)
-echo "Successfully Added managed identity to Function App '$functionAppName'"
+# add the reply url to the app registration reply urls
+az ad app update --id $clientId   --add replyUrls "https://${functionAppName}.azurewebsites.net/.auth/login/aad/callback"
```
manishsat: sudo mv microsoft.gpg /etc/apt/trusted.gpg.d/microsoft.gpg
this does not work on Mac

AhmedAbdoOrtiga: out of this PR scope but, it works for Ubuntu which is our target environment, it doesn't work for Windows or Mac, but if you're using some Ubunto distribution on your system like Ubuntu on Windows, it should work
look at this https://docs.microsoft.com/en-us/azure/azure-functions/functions-run-local#v2

packages/resource-deployment/scripts/function-app-create.sh
AhmedAbdoOrtiga: How?, id is required for "az ad app show", we can't use the name, and even if "az ad app show" support searching only by name,  it might not be unique, you can have two registration app with the same name

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -28,27 +27,55 @@ while getopts "r:" option; do
...
+    # Get the Client (App) ID for the App Registration that's used for this Function App authentication
+    clientId=$(az webapp auth show -n "$functionAppName" -g "$resourceGroupName" --query "clientId" -o tsv)
+    # Check if the App Registration name is valid and the App Registration exists
+    if [ ! -z "$clientId" ] && (az ad app show --id "$clientId" 1>/dev/null); then
+        # If we hit this line, that means we don't need to create a new App Registration
+        createNewApp=false
+    fi
+fi
+
+if ("$createNewApp" = true); then
```
haonliu:
```suggestion
if ($createNewApp = true); then
```

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -28,27 +27,55 @@ while getopts "r:" option; do
...
+    appRegistrationName="allyappregistration_${resourceGroupName}"
```
haonliu: appRegistrationName shouldn't be tied to the resource group?

packages/resource-deployment/scripts/function-app-create.sh
haonliu: Can make this part a function ``createAppRegistration()``? 
Then we don't have to deal with the boolean value which seems not supported in bash scripts?

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -28,27 +27,55 @@ while getopts "r:" option; do
...
-if [ -z "$resourceGroupName" ]; then
+if [ -z $resourceGroupName ]; then
     exitWithUsageInfo
 fi
 
+# Will return the function name if we deployed before the function app template on this resource group
+functionAppName=$(az group deployment show -g "$resourceGroupName" -n "function-app-template"  --query "properties.parameters.name.value" -o tsv)
+
+# Boolean to track if we should create a new App Registration or there's one already existing
+createNewApp=true
```
haonliu: this is just a string "true" right?

packages/resource-deployment/scripts/function-app-create.sh
devaradhanm: change it to allyfunctionApp_${environment}_${resourceGroupName}

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -28,27 +27,55 @@ while getopts "r:" option; do
...
 
-# Add system-assigned managed identity to function app
-
-echo "Adding managed identity to Function App '$functionAppName'"
-principalId=$(az webapp identity assign --name "$functionAppName" --resource-group "$resourceGroupName" --query principalId -o tsv)
-echo "Successfully Added managed identity to Function App '$functionAppName'"
+# Enable continue on error in case the redirect URl was already added before
+set +e
+# add the reply url to the app registration reply urls
+az ad app update --id $clientId --add replyUrls "https://${functionAppName}.azurewebsites.net/.auth/login/aad/callback"
```
devaradhanm: i dont get any error when reply url is already set.

packages/resource-deployment/scripts/function-app-create.sh
AhmedAbdoOrtiga: Yeah, but that's a way to handle Boolean in bash

packages/resource-deployment/scripts/function-app-create.sh
AhmedAbdoOrtiga: Add this line in a bash script then run it
image

packages/resource-deployment/scripts/function-app-create.sh
AhmedAbdoOrtiga: If we create a function, we'll move same logic to the function which still has boolean, unless you want to Not all the conditions in the if's and create assume that the default is we don't need to create a new one, which I don't like to do, I'd like to use the boolean as a guard for the knwon case that we don't need to create a new app

packages/resource-deployment/scripts/function-app-create.sh
AhmedAbdoOrtiga: I did it, but I'm not totally sold on it, if we want it to be also a standalone script, so the environment isn't related, maybe the subscription id.

packages/resource-deployment/scripts/function-app-create.sh
devaradhanm: you can still move it under createAppRegistration

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -7,48 +7,89 @@
...
     exitWithUsageInfo
 fi
 
-# Start deployment
+# Will return the function name if we deployed before the function app template on this resource group
+echo "Checking if the function app and AppRegistration already exists..."
+functionAppName=$(az group deployment show -g "$resourceGroupName" -n "function-app-template" --query "properties.parameters.name.value" -o tsv)
+clientId=$(az webapp auth show -n "$functionAppName" -g "$resourceGroupName" --query "clientId" -o tsv) || true
+
+if [[ ! -n $clientId ]]; then
```
AhmedAbdoOrtiga: this is not enough, we should still check if id is valid but the app is deleted
[ -z "$clientId" ] || (! az ad app show --id "$clientId" 1>/dev/null);

AhmedAbdoOrtiga: Also just wondering why not using the same format we use to check for null or empty which is 
[ -z ] instead of [ -n]

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -7,48 +7,89 @@
...
 . "${0%/*}/get-resource-name-from-resource-paths.sh" -p "Microsoft.Web/sites" -r "$resources"
 functionAppName=$resourceName
 echo "Successfully deployed Function App '$functionAppName'"
 
-# Add system-assigned managed identity to function app
-
-echo "Adding managed identity to Function App '$functionAppName'"
-principalId=$(az webapp identity assign --name "$functionAppName" --resource-group "$resourceGroupName" --query principalId -o tsv)
-echo "Successfully Added managed identity to Function App '$functionAppName'"
+addReplyUrlIfNotExist $clientId $functionAppName
```
devaradhanm: ifNotExists

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -7,48 +7,89 @@
...
-if [ -z "$resourceGroupName" ]; then
+if [ -z $resourceGroupName ] || [ -z $environment ]; then
     exitWithUsageInfo
 fi
 
-# Start deployment
+# Will return the function name if we deployed before the function app template on this resource group
+echo "Checking if the function app and AppRegistration already exists..."
+functionAppName=$(az group deployment show -g "$resourceGroupName" -n "function-app-template" --query "properties.parameters.name.value" -o tsv)
+clientId=$(az webapp auth show -n "$functionAppName" -g "$resourceGroupName" --query "clientId" -o tsv) || true
```
AhmedAbdoOrtiga: why do we need "|| true"

AhmedAbdoOrtiga: do we need to check first if the function name is valid and it still exists 
"[ ! -z "$functionAppName" ] && (az functionapp show -n "$functionAppName" -g "$resourceGroupName" 1>/dev/null)"
if it dowesn't, did we try if this is not gonna throw exception?

packages/resource-deployment/scripts/function-app-create.sh
haonliu: updated.

packages/resource-deployment/scripts/function-app-create.sh
haonliu: if the function app is deleted, clientId will be empty.

PR #271, opened 2019-09-06
Removing unused properties from products template

flyingUnderTheRadar:
#### Description of changes

#### Pull request checklist

### DIFF
--- a/packages/resource-deployment/templates/rest-api-templates/model-products.template.json
+++ b/packages/resource-deployment/templates/rest-api-templates/model-products.template.json
@@ -35,8 +35,6 @@
             "properties": {
                 "description": "Subscribers have completely unlimited access to the API. Administrator approval is required.",
                 "subscriptionRequired": false,
-                "approvalRequired": false,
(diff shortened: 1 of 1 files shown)


packages/resource-deployment/templates/rest-api-templates/model-products.template.json
```diff
@@ -35,8 +35,6 @@
             "properties": {
                 "description": "Subscribers have completely unlimited access to the API. Administrator approval is required.",
                 "subscriptionRequired": false,
-                "approvalRequired": false,
```
devaradhanm: what is the default value if approvalRequired is not set?

flyingUnderTheRadar: You can only approve if a subscription is required. Since a subscription is not required you cannot require a approval. In fact if subscriptionRequired is set to false then you can't specify the approvalRequired / subscription limit value at all. The approvalRequired setting is to say that each developer trying to subscribe to the product needs to be approved by the admin.

PR #278, opened 2019-09-10
Pass the app registration client ID to the function app

AhmedAbdoOrtiga:
#### Description of changes

If the client id is passed, then use it (pipeline)
if it's not, then create a new one (mostly to run it local ex: cmd)

#### Pull request checklist


packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -49,7 +51,7 @@ createAppRegistration() {
 
     appRegistrationName="allyappregistration-$resourceGroupName-$environment"
     echo "Creating a new AppRegistration with display name $appRegistrationName..."
-    clientId=$(az ad app create --display-name "$appRegistrationName" --query "appId" -o tsv)
+    clientId=$(az ad app create --display-name "$appRegistrationName" --query "appId" -o tsv) || true
```
haonliu: Should script exit if this line fails (which means ad app creation failed)?

AhmedAbdoOrtiga: No, I don't think, we should be able to handle that manually since it's a small change, we already do that in our release pipelines, we all can talk also if you think we should

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -28,7 +30,7 @@ addReplyUrlIfNotExists() {
 
     # Get existing reply urls of the app registration
     echo "Fetching existing replyUrls..."
-    replyUrls=$(az ad app show --id $clientId --query "replyUrls" -o tsv)
+    replyUrls=$(az ad app show --id $clientId --query "replyUrls" -o tsv) || true
```
devaradhanm: why is this change needed?

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -70,34 +72,37 @@ copyConfigFile() {
...
     e) environment=${OPTARG} ;;
     k) keyVault=${OPTARG} ;;
     d) dropFolder=${OPTARG} ;;
     *) exitWithUsageInfo ;;
     esac
 done
 
-if [ -z $resourceGroupName ] || [ -z $environment ] || [ -z $keyVault ]; then
+if [ -z $resourceGroupName ] || [ -z $clientId ] || [ -z $environment ] || [ -z $keyVault ]; then
     exitWithUsageInfo
```
devaradhanm: you shouldnt fail if clientId is not set. you should only fail if client id is not set & environment is not dev.

packages/resource-deployment/scripts/function-app-create.sh
AhmedAbdoOrtiga: because we still try to add in both cases, ClientId was passed or a new app was created, so we might need to face the same permission issue if we are running from a pipeline

packages/resource-deployment/scripts/function-app-create.sh
AhmedAbdoOrtiga: I'll fix it in another PR

PR #279, opened 2019-09-10
adding support to store scan run results

devaradhanm:
This PR adds library support to store/retrieve scan run results stored across partitions
(validated that querying with 1000 ids in a single query within a partition works)
#### Pull request checklist


packages/service-library/src/global.d.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import 'jest-extended';
```
lamaks: why not make it global import so any test can use it or use import within test file itself instead of hiding it in d.ts file?

packages/common/src/system/guid-utils.ts
```diff
@@ -8,21 +11,22 @@ import * as uuid from 'uuid-with-v6';
...
+    public createGuid(): string {
         return uuid.v6();
     }
 
     /**
      * Creates UUID with the same UUID node part (the last 12 bytes) as base UUID.
      * @param baseGuid The UUID to match the UUID node with.
      */
-    export function createGuidForNode(baseGuid: string): string {
+    public createGuidForNode(baseGuid: string): string {
```
manishsat: the name is bit confusing, it seems like we are creating Guid for Node. Instead you probably can change it to createNewGuidFromBaseGuid(); generateGuidFromBaseGuid() etc. And it will be clear to the caller what the function is going to do.

packages/common/src/system/guid-utils.ts
```diff
@@ -8,21 +11,22 @@ import * as uuid from 'uuid-with-v6';
 /**
  * UUID RFC 4122 https://tools.ietf.org/html/rfc4122
  */
-export namespace Guid {
+@injectable()
+export class GuidUtils {
```
lamaks: why not call it like guid-generator or provider or smth? Utils implies smth generic like a kitchen sink

packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+    public static readonly partitionKeyPreFix = 'pageScanRunResult';
+
+    constructor(
+        @inject(HashGenerator) private readonly hashGenerator: HashGenerator,
+        @inject(GuidUtils) private readonly guidUtils: GuidUtils,
+        @inject(cosmosContainerClientTypes.OnDemandScanRunsCosmosContainerClient)
+        private readonly cosmosContainerClient: CosmosContainerClient,
+    ) {}
+
+    public async createScanRuns(scanRuns: OnDemandPageScanResult[]): Promise<void> {
```
lamaks: this is actually to write documents can be writeScanRuns()

packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        await this.cosmosContainerClient.mergeOrWriteDocument(pageScanResult);
+    }
+
+    private getReadScanQueryForScanIds(scanIds: string[]): string {
+        const orConditions = scanIds.map(scanId => {
+            return `c.id = '${scanId}'`;
+        });
+
+        return `select * from c where ${orConditions.join(' or ')}`;
```
lamaks: The query has limits, if we change batch size this can be affected, we may do simple limit check here https://docs.microsoft.com/en-us/azure/cosmos-db/concepts-limits#sql-query-limits

packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+        });
+
+        await Promise.all(
+            Object.keys(scanRunsByPartition).map(async pKey => {
+                return this.cosmosContainerClient.writeDocuments(scanRunsByPartition[pKey], pKey);
+            }),
+        );
+    }
+
+    public async readScanRuns(scanIds: string[]): Promise<OnDemandPageScanResult[]> {
```
haonliu: Do we also need a method for reading a single scan run?

packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
haonliu: I'm wondering if we should use IN keyword instead. https://docs.microsoft.com/en-us/azure/cosmos-db/sql-query-keywords#in

packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
```diff
@@ -0,0 +1,81 @@
...
+        const scanIdsByPartition = groupBy(scanIds, scanId => {
+            return this.getPartitionKey(scanId);
+        });
+
+        const response = await Promise.all(
```
haonliu: What will happen if one promise call is somehow rejected?

packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+                scanIdsToFetch.push(`scanId-${i}`);
+            }
+
+            await expect(testSubject.readScanRuns(scanIdsToFetch)).rejects.toEqual(
+                new Error("Can't read more than 1000 scan documents per query"),
+            );
+        });
+
+        it('retrieves scan run documents', async () => {
+            const scanIdsToFetch = ['partition1id1', 'partition2id1', 'partition1id2'];
```
haonliu: Should we add a test case where the database query was rejected?

packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
devaradhanm: we will fail

PR #285, opened 2019-09-11
adding blob service to store/retrieve blob content

devaradhanm:
#### Description of changes

#### Pull request checklist


packages/azure-services/src/register-azure-services-to-container.ts
```diff
@@ -63,9 +65,20 @@ export function registerAzureServicesToContainer(container: Container): void {
...
+
     container.bind(Queue).toSelf();
 }
 
+function setupBlobServiceClientProvider(container: interfaces.Container): void {
+    IoC.setupSingletonProvider<BlobServiceClient>(iocTypeNames.BlobServiceClientProvider, container, async context => {
+        const secretProvider = context.container.get(SecretProvider);
+        const accountName = await secretProvider.getSecret(secretNames.storageAccountName);
+
+        return new BlobServiceClient(`https://${accountName}.blob.core.windows.net`, new DefaultAzureCredential());
```
lamaks: what data it has?

packages/azure-services/src/azure-blob/blob-service.ts
```diff
@@ -0,0 +1,55 @@
...
+import { isNil } from 'lodash';
+import { BlobServiceClientProvider, iocTypeNames } from '../ioc-types';
+
+export interface BlobContentDownloadResponse {
+    notFound: boolean;
+    content: NodeJS.ReadableStream;
+}
+
+@injectable()
+export class BlobService {
```
lamaks: can we call it client instead of service?

packages/azure-services/src/azure-blob/blob-service.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { BlobClient, BlobServiceClient, RestError } from '@azure/storage-blob';
```
lamaks: BlobServiceClient  - do we use this import?

PR #293, opened 2019-09-14
Web API Job Manager and Runner

AhmedAbdoOrtiga:
#### Description of changes
Web API Job Manager and Runner

#### Pull request checklist


packages/resource-deployment/scripts/upload-files.sh
```diff
@@ -6,7 +6,9 @@
 set -eo pipefail
 
 export jobManagerContainerName="batch-job-manager-script"
+export webAPIScanJobManagerContainerName="batch-web-api-scan-job-manager-script"
```
devaradhanm: >webAPIScanJobManagerContainerName [](start = 7, length = 33)

webApi

AhmedAbdoOrtiga: API isn't a word, it's initials, so I think it should be API

packages/web-api-scan-job-manager/run-script/run-web-api-scan-job-manager.sh
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+#!/bin/bash
+export NODE_PATH=$AZ_BATCH_NODE_SHARED_DIR/batch-job-manager/node_modules
```
devaradhanm: >batch-job-manage [](start = 43, length = 16)

this path has to be updated

packages/web-api-scan-job-manager/src/batch/batch-config.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT License.
+import 'reflect-metadata';
+
+import { BatchConfig } from './batch-config';
+
+describe(BatchConfig, () => {
+    let batchConfig: BatchConfig;
+
+    it('return value of AZ_BATCH_ACCOUNT_NAME environment variable', () => {
+        const value: string = `value-${new Date().valueOf()}`;
```
devaradhanm: >`value-${new Date().valueOf()} [](start = 30, length = 30)

this could be a constant value.
also you could just have 1 unit test that validates all 3 properties.:
process.env.AZ_BATCH_ACCOUNT_NAME = "test batch account1";
process.env.AZ_BATCH_ACCOUNT_URL = "test account url1";
...

expect(batchConfig.accountName).toEqual(process.env.AZ_BATCH_ACCOUNT_NAME);
...

packages/web-api-scan-job-manager/src/batch/job-task.spec.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { JobTask } from './job-task';
+
+describe(JobTask, () => {
+    let jobTask: JobTask;
+
+    it('id should include correlation value', () => {
+        const correlationId = `correlationId_${new Date().valueOf()}`;
```
devaradhanm: >{new Date().valueOf()} [](start = 46, length = 22)

avoid using changing values for test

AhmedAbdoOrtiga: this is the same in the old package, I'd like to focus on what was changed first for the sake of time

packages/web-api-scan-job-manager/src/batch/batch-config.spec.ts
AhmedAbdoOrtiga: this is the same in the old package, I'd like to focus on what was changed first for the sake of time

packages/resource-deployment/scripts/upload-files.sh
devaradhanm: i know. but when you use this in a variable name, we don't do that, so that it is easier to know where the word ends. so, it should be webApiScanJobManagerContainerName

packages/web-api-scan-job-manager/package.json
```diff
@@ -3,7 +3,7 @@
     "version": "1.0.0",
     "description": "This project welcomes contributions and suggestions.  Most contributions require you to agree to a Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us the rights to use your contribution. For details, visit https://cla.microsoft.com.",
     "scripts": {
-        "build": "webpack --config ./webpack.config.js --colors",
+        "build": "webpack --config ./webpack.config.js --config-name web-api-scan-job-manager --colors",
```
devaradhanm: this parameter is not used if the webpack config doesnt contain array of configuration

packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -0,0 +1,88 @@
...
+        const scanMetadata = this.scanMetadataConfig.getConfig();
+        let onDemandScanResult: OnDemandScanResult;
+        let onDemandPageScanReport: OnDemandPageScanReport;
+        let onDemandPageScanRunResult: OnDemandPageScanRunResult;
+
+        try {
+            const pageScanRunResult = (await this.onDemandPageScanRunResultProvider.readScanRuns([scanMetadata.id]))[0];
+
+            onDemandScanResult = {
+                state: 'unknown' as ScanState,
```
devaradhanm: you shouldnt be casting these types.

packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -0,0 +1,88 @@
...
+                issueCount: 0,
```
devaradhanm: set issueCount to undefined.

packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -0,0 +1,88 @@
...
+                .toString();
+
+            onDemandScanResult.issueCount = axeScanResults.results.violations.length;
+            onDemandScanResult.state = (onDemandScanResult.issueCount > 0 ? 'fail' : 'pass') as ScanState;
+
+            pageScanRunResult.run = onDemandPageScanRunResult;
+            pageScanRunResult.reports = [onDemandPageScanReport];
+            pageScanRunResult.scanResult = onDemandScanResult;
+
+            await this.onDemandPageScanRunResultProvider.updateScanRun(pageScanRunResult);
```
devaradhanm: am not sure if this will work. etag value will be different now, since you have already updated this object with a etag value. if you modify the object & call update api, since you are still having the old etag value it may not work.

AhmedAbdoOrtiga: As discussed offline, it should be fine

packages/resource-deployment/scripts/upload-files.sh
AhmedAbdoOrtiga: let's discuss offline

packages/web-api-scan-job-manager/package.json
AhmedAbdoOrtiga: we still add it in other packages

devaradhanm: yes. but it is wrong. I have fixed it in some places. but not everywhere. you don't have to carry over the issue since we know that it doesnt work.

flyingUnderTheRadar:
If the PR is in its final form can you pull from master please?

packages/web-api-scan-job-manager/src/worker/worker.ts
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            await System.wait(this.jobManagerConfig.addTasksIntervalInSeconds * 1000);
+        }
+    }
+
+    private async getMessages(messagesCount: number): Promise<Message[]> {
+        const messages: Message[] = [];
+        do {
+            const batch = await this.queue.getMessages();
```
flyingUnderTheRadar: can we rename this to queuedMessages / deQueuedMessages or something similar to be clearer about the value in this const?

packages/web-api-scan-job-manager/src/worker/worker.ts
```diff
@@ -0,0 +1,114 @@
...
+            }),
+        );
+
+        const jobQueuedTasks = jobTasks.filter(jobTask => jobTask.state === JobTaskState.queued);
+
+        return jobQueuedTasks.length;
+    }
+
+    private async init(): Promise<void> {
+        this.jobManagerConfig = await this.serviceConfig.getConfigValue('jobManagerConfig');
```
flyingUnderTheRadar: Is this config a shared config with the other job manager?

packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -0,0 +1,104 @@
...
+import { inject, injectable } from 'inversify';
+import { Logger } from 'logger';
+import { Browser } from 'puppeteer';
+import { AxeScanResults } from 'scanner';
+import { OnDemandPageScanRunResultProvider } from 'service-library';
+import {
+    OnDemandPageScanReport,
+    OnDemandPageScanResult,
+    OnDemandPageScanRunResult,
+    OnDemandPageScanRunState,
```
flyingUnderTheRadar: Can we remove unused imports please?

packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -0,0 +1,104 @@
...
+            onDemandPageScanReport = {
+                reportId: '',
+                format: 'sarif',
+                href: '',
+            };
+
+            // set scanned page run state to running
+            onDemandPageScanRunResult = {
+                state: 'running',
+                timestamp: new Date()
```
flyingUnderTheRadar: `new Date().ToJson()` will suffice.

packages/web-api-scan-job-manager/package.json
AhmedAbdoOrtiga: how doesn't it work? can you elaborate?

packages/web-api-scan-job-manager/src/worker/worker.ts
AhmedAbdoOrtiga: I think the name makes sense cause we only read the messages, and later we remove it, so it's not dequeued yet

packages/web-api-scan-job-manager/src/worker/worker.ts
AhmedAbdoOrtiga: yep, it should be the same config

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -87,7 +87,7 @@ if [ -z $resourceGroupName ] || [ -z $environment ] || [ -z $keyVault ]; then
     exitWithUsageInfo
 fi
 
-if [ -z $clientId] && [ ! "${environment,,}" = "dev" ]; then
+if [ -z $clientId ] && [ ! "${environment,,}" = "dev" ]; then
```
flyingUnderTheRadar: Do we know why we have the `,,` ?

packages/web-api-scan-job-manager/src/worker/worker.ts
flyingUnderTheRadar: Honestly, `batch` does not tell me anything about what the variable is holding. What is it a batch of ? We don't have the concept of batch at all. Some other options you might want to consider - `fetchedMessages` `scanRequestMessages` `messagesToScan` `scanRequests`. All of these tell you something about what the variable is holding.

packages/resource-deployment/scripts/function-app-create.sh
AhmedAbdoOrtiga: to lower case

packages/resource-deployment/templates/blob-storage.template.json
```diff
@@ -64,6 +71,13 @@
         "runnerContainerName": {
             "type": "string",
             "defaultValue": "batch-runner-script",
+            "metadata": {
+                "description": "The name of the blob container for web api scan runner script."
+            }
+        },
+        "webAPIScanRunnerContainerName": {
```
devaradhanm: as discussed offline, webApiScanRunnerContainerName

packages/resource-deployment/templates/url-web-api-scan-schedule.template.json
```diff
@@ -0,0 +1,59 @@
+{
+    "id": "url-web-api-scan-schedule",
+    "schedule": {
+        "recurrenceInterval": "PT10M"
```
devaradhanm: change this to 1 min

packages/web-api-scan-job-manager/src/batch/runner-task-config.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+                KEY_VAULT_URL: keyVaultUrl,
+                RUNNER_SCRIPTS_CONTAINER_NAME: runnerScriptContainerName,
+            } as NodeJS.ProcessEnv,
+        } as NodeJS.Process;
+
+        testSubject = new RunnerTaskConfig(processStub);
+    });
+
+    it('builds command line', () => {
+        const data = { id: 'id', url: 'url', priority: 1 };
```
devaradhanm: have value like id1, url1

packages/web-api-scan-job-manager/src/setup-job-manager-container.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    beforeEach(() => {
+        process.env.AZURE_STORAGE_SCAN_QUEUE = 'test-scan-queue';
+
+        process.env.AZ_BATCH_ACCOUNT_NAME = batchAccountName;
+        process.env.AZ_BATCH_ACCOUNT_URL = batchAccountUrl;
+        process.env.AZ_BATCH_POOL_ID = 'test-batch-pool-id';
+    });
+
+    describe('BatchServiceClient', () => {
```
devaradhanm: remove this decribe layer & move content under it to parent describe

packages/web-api-scan-job-manager/src/worker/worker.spec.ts
```diff
@@ -0,0 +1,244 @@
...
+        batchMock = Mock.ofType(Batch);
+        queueMock = Mock.ofType(Queue);
+        poolLoadGeneratorMock = Mock.ofType(PoolLoadGenerator);
+        serviceConfigMock = Mock.ofType(ServiceConfiguration);
+        loggerMock = Mock.ofType(Logger);
+
+        queueMock.setup(o => o.scanQueue).returns(() => 'scan-queue');
+
+        serviceConfigMock
+            .setup(async s => s.getConfigValue('jobManagerConfig'))
```
devaradhanm: you should probable create another config instead of using jobManagerConfig
both jobManagerConfig & onDemandJobManager can be of same type

devaradhanm: or u could do this in next pr.

packages/web-api-scan-runner/src/runner/runner.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+let browser: Browser;
+let webDriverTaskMock: IMock<WebDriverTask>;
+let onDemandPageScanRunResultProviderMock: IMock<OnDemandPageScanRunResultProvider>;
+let scannerTaskMock: IMock<ScannerTask>;
+let scanMetadataConfig: IMock<ScanMetadataConfig>;
+let loggerMock: IMock<Logger>;
+let pageScanRunReportServiceMock: IMock<PageScanRunReportService>;
+let guidGeneratorMock: IMock<GuidGenerator>;
+
+const scanMetadata: ScanMetadata = {
```
devaradhanm: you need to rewrite the tests for this:
test cases: 
1.) when axeResult returns error
2.) when axeResult return with violations length  = 0
3.) when axeResult returns with violations length > 0
4.) web driver launch crashes
5.) web driver close crashes
6.) scanner task crashes
7.) cosmos read/write operation crashes

PR #298, opened 2019-09-17
get report rest api

haonliu:
#### Description of changes

#### Pull request checklist


packages/web-api/src/controllers/scan-report-controller.ts
```diff
@@ -0,0 +1,50 @@
...
+    ) {
+        super();
+    }
+
+    public async handleRequest(): Promise<void> {
+        const reportId = <string>this.context.bindingData.reportId;
+        if (!this.guidGenerator.isValidV6Guid(reportId)) {
+            this.context.res = {
+                status: 422,
+                body: `Unprocessable Entity: ${reportId}.`,
```
manishsat: Change to "invalid report id" or something.

haonliu: done.

packages/web-api/src/controllers/scan-report-controller.ts
```diff
@@ -0,0 +1,54 @@
...
+        if (blobContentDownloadResponse.notFound === true) {
+            this.context.res = {
+                status: 404,
+            };
+
+            return;
+        }
+
+        const stream = new Stream.PassThrough();
+        blobContentDownloadResponse.content.pipe(stream);
```
devaradhanm: can't we directly pass the stream returned by blobContentDownloadResponse.content?

devaradhanm: Or is this the pattern when dealing with rest api response?

flyingUnderTheRadar: With [other frameworks](https://dev.azure.com/mseng/_git/AzureDevOps?path=%2FKeros%2FService%2FServer%2FUtility%2FHttpResponseMessageCreator.cs&version=GBmaster) it is possible to send a stream. Not sure whether we can do it with typescript. We can also punt this to a bug.

haonliu: I passed content as body directly initially but the client never get response.

packages/web-api/src/controllers/scan-report-controller.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        if (blobContentDownloadResponse.notFound === true) {
+            this.context.res = {
+                status: 404,
+            };
+
+            return;
+        }
+
+        const content = await getRawBody(blobContentDownloadResponse.content as Readable);
```
devaradhanm: have you validated if this works & u are able to read the content?

PR #300, opened 2019-09-17
Waiting till all running tasks are done.

manishsat:
#### Description of changes

Job manager will exit only if there are no active and running tasks.

#### Pull request checklist

### DIFF
--- a/packages/job-manager/src/worker/worker.ts
+++ b/packages/job-manager/src/worker/worker.ts
@@ -38,8 +38,8 @@ export class Worker {
                 const scanMessages = await this.getMessages(tasksIncrementCount);
                 if (scanMessages.length === 0) {
                     this.logger.logInfo(`The storage queue '${this.queue.scanQueue}' has no message to process.`);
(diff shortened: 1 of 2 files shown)


packages/job-manager/src/worker/worker.ts
```diff
@@ -38,8 +38,8 @@ export class Worker {
                 const scanMessages = await this.getMessages(tasksIncrementCount);
                 if (scanMessages.length === 0) {
                     this.logger.logInfo(`The storage queue '${this.queue.scanQueue}' has no message to process.`);
-                    if (poolMetricsInfo.load.activeTasks === 0) {
-                        this.logger.logInfo(`Exiting the ${this.jobId} job since there are no active tasks.`);
+                    if (poolMetricsInfo.load.activeTasks === 0 && poolMetricsInfo.load.runningTasks === 0) {
```
devaradhanm: uts missing

devaradhanm: (u should have a test case where active tasks is > 0 & running tasks =  0 & vise versa

PR #312, opened 2019-09-20
Retry func tool deployment run

lamaks:
#### Description of changes

Retry func tool deployment run

#### Pull request checklist

### DIFF
--- a/packages/resource-deployment/scripts/function-app-create.sh
+++ b/packages/resource-deployment/scripts/function-app-create.sh
@@ -134,7 +134,12 @@ publishFunctionAppScripts() {
         sleep 5
     done
 
-    echo "Successfully published API functions to '$currentFunctionAppName' Function App."
+    if [ $? -ne 0 ]; then
+        echo "Publishing '$packageName' scripts to '$currentFunctionAppName' Function App was unsuccessful."
+        exit 1
+    fi
+
+    echo "Successfully published '$packageName' scripts to '$currentFunctionAppName' Function App."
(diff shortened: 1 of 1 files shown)


packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -123,9 +123,18 @@ publishFunctionAppScripts() {
 
     # Publish the function scripts to the function app
     echo "Publishing '$packageName' scripts to '$currentFunctionAppName' Function App..."
-    func azure functionapp publish $currentFunctionAppName --node
-    echo "Successfully published API functions to '$currentFunctionAppName' Function App."
 
+    # Run function tool with retries due to app service warm up time delay
+    end=$((SECONDS + 120))
+    while [ $SECONDS -le $end ]; do
+        func azure functionapp publish $currentFunctionAppName --node || true
```
devaradhanm: can you print if the func publish fails? in this case, if the func publish fails after timeout, you still consider it to be successful

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -123,9 +123,18 @@ publishFunctionAppScripts() {
...
+        if [ $? -eq 0 ]; then
```
devaradhanm: can you print if the func publish fails? in this case, if the func publish fails after timeout, you still consider it to be successful

lamaks: updated

PR #313, opened 2019-09-20
Perf improvement for fresh install and repeat install (20 - 45%)

flyingUnderTheRadar:
#### Description of changes

Perf improvements of ~20 & ~45% (repeat install & clean install)

Runs all the scripts that have the same dependencies in parallel. 

The install script is purposely in a different file where we can test it for a couple of days following which we can replace the install file.

Sample runs (3 each):
- Fresh install
-- Sequential: 54 - 56 mins
-- Parallel: 30 - 34 mins
- Same infra install
-- Sequential: 17 - 20 mins
-- Parallel: 13 - 16 mins
#### Pull request checklist


packages/resource-deployment/scripts/install-parallel.sh
```diff
@@ -0,0 +1,145 @@
...
+    exit 1
+}
+
+
+function waitForProcesses(){
+    processesToWaitFor=$1
+    for pid in ${processesToWaitFor[@]}; do
+        wait $pid
+        returnCode=$?
+        echo "Value of pid is $pid, Return code is $returnCode"
```
flyingUnderTheRadar: this doesn't add any value beyond getting us the return codes for the background processes.

packages/resource-deployment/scripts/install-parallel.sh
```diff
@@ -0,0 +1,145 @@
...
+
+. "${0%/*}/create-storage-account.sh"
+
+echo "Finished creating storage account"
+
+resourceGroupSuffix=${storageAccountName:11}
+cosmosAccountName="allycosmos$resourceGroupSuffix"
+apiManagementName="apim-a11y$resourceGroupSuffix"
+
+. "${0%/*}/create-api-management.sh" &
```
devaradhanm: it may not be a good idea to source the script when you are parallelizing them. it may have unpredictable state on the parent process' environment variables.

devaradhanm: if we are planning to parallelize, we may need to refactor out the code that sets environment variables out & not parallelize that part.

flyingUnderTheRadar: background process cannot touch the parents variable at all !

packages/resource-deployment/scripts/install-parallel.sh
```diff
@@ -0,0 +1,145 @@
...
+
+. "${0%/*}/batch-account-create.sh"
+
+. "${0%/*}/push-secrets-to-key-vault.sh"
+
+. "${0%/*}/function-app-create.sh"
+
+. "${0%/*}/job-schedule-create.sh"
+
+waitForProcesses "$apiManagmentProcess"
```
lamaks: We are waiting for all processes anyway and could follow same pattern by adding pid into array

packages/resource-deployment/scripts/install-parallel.sh
```diff
@@ -0,0 +1,145 @@
...
+echo "Finished creating storage account"
+
+resourceGroupSuffix=${storageAccountName:11}
+cosmosAccountName="allycosmos$resourceGroupSuffix"
+apiManagementName="apim-a11y$resourceGroupSuffix"
+
+. "${0%/*}/create-api-management.sh" &
+apiManagmentProcess="$!"
+
+. "${0%/*}/create-datalake-storage-account.sh" &
```
lamaks: Can we move all background scripts into func to separate logic

packages/resource-deployment/scripts/install-parallel.sh
```diff
@@ -0,0 +1,145 @@
...
+    francecentral
+    southafricanorth
+    uaenorth
+"
+    exit 1
+}
+
+
+function waitForProcesses(){
+    processesToWaitFor=$1
```
lamaks: Not sure we need script exit code, we can drop return/echo and just wait for processes. Will wait fail if process already was terminated before we call wait for that process id? We may change the wait call then similar to: 

wait $pid 2>/dev/null || true

packages/resource-deployment/scripts/install-parallel.sh
```diff
@@ -0,0 +1,145 @@
...
+    az login
+fi
+
+az account set --subscription "$subscription"
+
+. "${0%/*}/create-resource-group.sh"
+
+. "${0%/*}/create-storage-account.sh"
+
+echo "Finished creating storage account"
```
lamaks: We can remove echo or move it to the storage script

packages/resource-deployment/scripts/create-api-management.sh
```diff
@@ -38,7 +38,7 @@ if [[ -z $orgName ]] || [[ -z $publisherEmail ]] || [[ -z $resourceGroupName ]];
 fi
 
 # Start deployment
-echo "Deploying API management instance. This might take up to 45 mins"
+echo "[create-api-management] Deploying API management instance. This might take up to 45 mins"
```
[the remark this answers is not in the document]
flyingUnderTheRadar: When run in a ci / cd env timestamps are auto added. Do we still want to add them explicitly?

packages/resource-deployment/scripts/install-parallel.sh
flyingUnderTheRadar: Some of these need to run sequentially because the predecessor sets some variables that are used in the successor. I haven't found a way to have a background process set the parent's variables. Plus it would be a race condition if we chose to run all in parallel.

lamaks: Can we have a comment of sequences since this is not obvious from the first look or what depends on what?

flyingUnderTheRadar: Discussed offline. Adding a comment to indicate dependency.

packages/resource-deployment/scripts/install-parallel.sh
flyingUnderTheRadar: Done. #Pending

packages/resource-deployment/scripts/create-api-management.sh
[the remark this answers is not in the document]
flyingUnderTheRadar: Discussed offline.

packages/resource-deployment/scripts/install-parallel.sh
flyingUnderTheRadar: Discussed offline. #NoAction

packages/resource-deployment/scripts/install-parallel.sh
flyingUnderTheRadar: Discussed offline. #No Action

PR #324, opened 2019-09-23
Increase too soon request time and change 'unknown' state to 'not found'

AhmedAbdoOrtiga:
#### Description of changes

- Increase too soon request time.
- Change 'unknown' state to 'not found'.

#### Pull request checklist


packages/storage-documents/src/on-demand-page-scan-result.ts
```diff
@@ -5,7 +5,7 @@ import { ItemType } from './item-type';
 
 export declare type ReportFormat = 'sarif';
 export declare type ScanState = 'unknown' | 'pass' | 'fail';
-export declare type OnDemandPageScanRunState = 'unknown' | 'accepted' | 'queued' | 'running' | 'completed' | 'failed';
+export declare type OnDemandPageScanRunState = 'accepted' | 'queued' | 'running' | 'completed' | 'failed';
```
devaradhanm: you also need to fix the get404ErrorResponse method in azure function. that is returning unknown value for run.

packages/common/src/configuration/__snapshots__/service-configuration.spec.ts.snap
```diff
@@ -45,7 +45,7 @@ Object {
       "format": "int",
     },
     "minimumWaitTimeforScanResultQueryInSeconds": Object {
-      "default": 120,
+      "default": 300,
```
devaradhanm: don't increase this timeout

packages/common/src/configuration/service-configuration.ts
```diff
@@ -167,7 +167,7 @@ export class ServiceConfiguration {
                 },
                 minimumWaitTimeforScanResultQueryInSeconds: {
                     format: 'int',
-                    default: 120,
+                    default: 300,
```
lamaks: In DEV we may need short time. Can we set default to 15 sec and change config files instead?

AhmedAbdoOrtiga: what config files, do we have different config file per profile? let's discuss it

packages/common/src/configuration/__snapshots__/service-configuration.spec.ts.snap
```diff
@@ -166,7 +166,7 @@ Object {
       "format": "int",
     },
     "minimumWaitTimeforScanResultQueryInSeconds": Object {
```
devaradhanm: can u also remove unknown from scan status. let scan property be null, if we don;t have the info

AhmedAbdoOrtiga: how, we should we put if there's no info?

devaradhanm: ig there is no info set it to null instead of scanstatus unknown

devaradhanm: but this could be in a different pr

packages/common/src/configuration/service-configuration.ts
lamaks: Set the value to 15 seconds to support development
Update config files here https://github.com/microsoft/accessibility-insights-service/tree/master/packages/resource-deployment/runtime-config

packages/storage-documents/src/on-demand-page-scan-result.ts
```diff
@@ -5,7 +5,7 @@ import { ItemType } from './item-type';
 
 export declare type ReportFormat = 'sarif';
 export declare type ScanState = 'unknown' | 'pass' | 'fail';
-export declare type OnDemandPageScanRunState = 'unknown' | 'accepted' | 'queued' | 'running' | 'completed' | 'failed';
+export declare type OnDemandPageScanRunState = 'pending' | 'not found' | 'accepted' | 'queued' | 'running' | 'completed' | 'failed';
```
lamaks: Please revert all changes related to 'pending' This is will included in my PR

AhmedAbdoOrtiga: There's a lot of confusion been happening, in the meeting I was asked to fix this, not sure why you will do it or what's the point of removing such a small change so you can repeat it, but I'll do it for the sake of getting this closed because the review it taking way more time than it should be.

lamaks: As per our offline conversation yesterday I  am working on web api controllers rest api response codes and scan run result format fix. This will have changes related to all web api controllers and include state result update. This is what I was meant by asking to not do any changes related to state yesterday. Can you please update only time (see above open comment) for 'too late response'. Thank you

AhmedAbdoOrtiga: As per our offline conversation, I told you I already did it after you asked because I was asked in the meeting to do it, and I told feel free to add what you want to the "PR", there's a lot of confusion needs to be addressed before I fix anymore comments

PR #326, opened 2019-09-24
Setup throughput based on environment.

manishsat:
#### Description of changes

Setup throughput based on environment, higher throughput in prod.

#### Pull request checklist

### DIFF
--- a/packages/resource-deployment/scripts/setup-cosmos-db.sh
+++ b/packages/resource-deployment/scripts/setup-cosmos-db.sh
@@ -25,6 +25,7 @@ createCosmosCollection() {
     local collectionName=$1
     local dbName=$2
     local ttl=$3
+    local throughput=$4
 
     if [[ -z $ttl ]]; then
         ttl=-1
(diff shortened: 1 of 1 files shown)


packages/resource-deployment/scripts/setup-cosmos-db.sh
```diff
@@ -85,8 +89,17 @@ onDemandScannerDbName="onDemandScanner"
 createCosmosDatabase "$scannerDbName"
 createCosmosDatabase "$onDemandScannerDbName"
 
-createCosmosCollection "a11yIssues" "$scannerDbName"
+# Increase throughput for below collection only in case of prod
 # Refer to https://docs.microsoft.com/en-us/azure/cosmos-db/time-to-live for item TTL scenarios
-createCosmosCollection "scanRuns" "$onDemandScannerDbName" "2592000"         # 30 days
-createCosmosCollection "scanBatchRequests" "$onDemandScannerDbName" "604800" # 7 days
-createCosmosCollection "scanRequests" "$onDemandScannerDbName" "604800"      # 7 days
+if [ $environment = "prod" ]; then
```
lamaks: Just to remove container/db name duplication can we move throughput to var and set in 'if' condition instead

packages/resource-deployment/scripts/setup-cosmos-db.sh
```diff
@@ -85,8 +89,17 @@ onDemandScannerDbName="onDemandScanner"
...
+    createCosmosCollection "a11yIssues" "$scannerDbName" "-1" "25000"
```
devaradhanm: how are u creating the cosmos collections if not prod?

packages/resource-deployment/scripts/setup-cosmos-db.sh
manishsat: I have considered that option too, which says set the throughput to a default value and change if the default value only if env is prod. But are we ok to consider single throughput for all collections?

PR #328, opened 2019-09-26
storing priority in batch request document;

devaradhanm:
not storing requests that won't have scan id

#### Description of changes

#### Pull request checklist

### DIFF
--- a/packages/web-api/src/controllers/scan-request-controller.ts
+++ b/packages/web-api/src/controllers/scan-request-controller.ts
@@ -2,8 +2,10 @@
 // Licensed under the MIT License.
 import { GuidGenerator, ServiceConfiguration, Url } from 'common';
 import { inject, injectable } from 'inversify';
+import { isNil } from 'lodash';
 import { Logger } from 'logger';
 import { ApiController } from 'service-library';
+import { ScanRunBatchRequest } from 'storage-documents';
 import { ScanRunRequest } from '../api-contracts/scan-run-request';
 import { ScanRunResponse } from '../api-contracts/scan-run-response';
 import { ScanDataProvider } from '../providers/scan-data-provider';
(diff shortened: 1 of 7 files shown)


packages/web-api/src/controllers/scan-request-controller.ts
```diff
@@ -39,36 +41,56 @@ export class ScanRequestController extends ApiController {
...
-            totalUrls: response.length.toString(),
-            invalidUrls: response.filter(i => i.error !== undefined).length.toString(),
+            totalUrls: processedData.responseToBeSent.length.toString(),
+            invalidUrls: processedData.responseToBeSent.filter(i => i.error !== undefined).length.toString(),
         });
     }
 
-    private createScanRunBatchResponse(batchId: string, scanRunRequests: ScanRunRequest[]): ScanRunResponse[] {
-        return scanRunRequests.map(scanRunRequest => {
+    private getProcessedRequestData(batchId: string, scanRunRequests: ScanRunRequest[]): ProcessedBatchRequestData {
```
manishsat: Function name is little vague, the name of the function should convey what it does.
In this case it determines the scanId(s) for the url(s), or createScantId(s) or generateScanId(s) or assignScanId(s)

devaradhanm: it is doing more than creating scan id. i named it similar to the return object type

packages/web-api/src/controllers/scan-request-controller.ts
```diff
@@ -39,36 +41,56 @@ export class ScanRequestController extends ApiController {
...
+
+        return {
+            requestToBeSaved,
+            responseToBeSent,
+        };
     }
 }
+
+interface ProcessedBatchRequestData {
+    requestToBeSaved: ScanRunBatchRequest[];
```
lamaks: The properties name is vague. Can be scanRunBatchResponseDocuments & scanRunClientResponses

devaradhanm: the two objects are not response objects. 1 of them is response object & the other is the request that will be stored

devaradhanm: I also wanted to rename ScanRunBatchRequest to ScanRunRequest, since the type itself is not a batch. But, thought of not doing it as part of this PR.

PR #330, opened 2019-09-26
Fix issue count

AhmedAbdoOrtiga:
#### Description of changes

- Go through all violations results and count how many nodes failed for same rule.

#### Pull request checklist

### DIFF
--- a/packages/web-api-scan-runner/src/runner/runner.ts
+++ b/packages/web-api-scan-runner/src/runner/runner.ts
@@ -113,7 +113,7 @@ export class Runner {
         let state: ScanState = 'fail';
 
         if (axeResults.results.violations !== undefined && axeResults.results.violations.length > 0) {
-            issueCount = axeResults.results.violations.length;
+            issueCount = axeResults.results.violations.reduce((a, b) => a + b.nodes.length, 0);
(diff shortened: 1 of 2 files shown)


packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -113,7 +113,7 @@ export class Runner {
         let state: ScanState = 'fail';
 
         if (axeResults.results.violations !== undefined && axeResults.results.violations.length > 0) {
-            issueCount = axeResults.results.violations.length;
+            issueCount = axeResults.results.violations.reduce((a, b) => a + b.nodes.length, 0);
```
devaradhanm: you need to add a proper unit test that validates that you are adding nodes length across violation rules objects

devaradhanm: you need to add a proper unit test that validates that you are adding nodes length across violation rules objects

AhmedAbdoOrtiga: fixed, but what do you mean by proper, If you looked at the UT file, it was modified to match the new change already and what we added now is just to make clear not proper

PR #337, opened 2019-09-27
Web REST API client response codes

lamaks:
#### Description of changes

Web REST API client response codes

#### Pull request checklist


packages/service-library/src/web-api/web-api-error-codes.ts
```diff
@@ -0,0 +1,106 @@
...
+    | 'InvalidResourceId'
+    | 'InvalidJsonDocument'
+    | 'RequestBodyTooLarge'
+    | 'InvalidURL'
+    | 'InternalError'
+    | 'MissingApiVersionQueryParameter'
+    | 'MissingContentTypeHeader'
+    | 'UnsupportedContentType'
+    | 'UnsupportedApiVersion';
+
```
devaradhanm: this change looks like it overlaps with the error codes feature. can you also address these if you are planning to cover that as part of this PR? 

1.) Invalid url
2.) Invalid page type (not a html page. Could be json/xml/image/etc url)
2.) Invalid input (missing required props, invalid json etc)
3.) Puppeteer crash
4.) timeout on loading
5) for other internal service issues like storage access error, queue access etc, we will have 1 error code - Internal Service Error
6.) For redirect url, we will not return error. Instead, the scan will be successful & return info about the redirected url with scan summary
7.) if failing for language issue, we need to see if we can identify this & have an error code for this. - so far, we have confirmed that we don't have language issues.
8.) axe scan error

packages/service-library/src/web-api/web-api-error-codes.ts
```diff
@@ -0,0 +1,106 @@
...
+    public static invalidResourceId: WebApiErrorCode = {
+        statusCode: 400,
+        response: {
+            code: 'InvalidResourceId',
+            error: 'The resource ID is not valid.',
+        },
+    };
+
+    public static invalidURL: WebApiErrorCode = {
+        statusCode: 400,
```
devaradhanm: do you also want to add http status code for this?
what about batch requests? you can have 1 url that is not valid, in that case are we planning to reject the whole request or just for the invalid ones? same applies for the internal error error code.

lamaks: In a batch we will return scan result object in array w/o rejecting request

packages/service-library/src/web-api/web-api-error-codes.ts
lamaks: The current PR covers only HTTP responses w/o detecting backend failures

PR #339, opened 2019-09-27
Incorporate Web REST API HTTP response codes

lamaks:
#### Description of changes

Incorporate Web REST API HTTP response codes.
Update scan run REST API client response.
Delete unused code.

#### Pull request checklist


packages/azure-services/src/register-azure-services-to-container.ts
```diff
@@ -79,9 +78,16 @@ export function registerAzureServicesToContainer(container: Container, credentia
...
+        let accountName: string;
+        let accountKey: string;
+        if (process.env.AZURE_STORAGE_NAME !== undefined && process.env.AZURE_STORAGE_KEY !== undefined) {
+            accountName = process.env.AZURE_STORAGE_NAME;
+            accountKey = process.env.AZURE_STORAGE_KEY;
+        } else {
+            const secretProvider = context.container.get(SecretProvider);
+            accountName = await secretProvider.getSecret(secretNames.storageAccountName);
+            accountKey = await secretProvider.getSecret(secretNames.storageAccountKey);
+        }
```
manishsat: This block seems to be glue code to me, setupSingletonQueueServiceURLProvider() method is also doing the same thing. Can we move this to a common place?

packages/web-api/src/controllers/batch-scan-result-controller.ts
```diff
@@ -26,9 +27,13 @@ export class BatchScanResultController extends BaseScanResultController {
 
     public async handleRequest(): Promise<void> {
         const payload = this.tryGetPayload<ScanBatchRequest[]>();
+        if (payload === undefined) {
+            return;
```
manishsat: Is this going to return 200?

packages/web-api/src/controllers/batch-scan-result-controller.ts
```diff
@@ -26,9 +27,13 @@ export class BatchScanResultController extends BaseScanResultController {
...
+        }
+
         if (isEmpty(payload)) {
             this.context.res = {
-                status: 422, // Unprocessable Entity,
+                status: 204, // No Content
```
manishsat: This if and the one just above it, can be moved to WebController? I am assuming we need to do this for all type of APIController.

packages/web-api/src/controllers/batch-scan-result-controller.ts
lamaks: Will return 400 InvalidJsonDocument

packages/web-api/src/api-contracts/scan-result-response.ts
```diff
@@ -20,6 +17,12 @@ export type ScanResultResponse =
       }
     | ScanResultErrorResponse;
 
+export interface ScanResultErrorResponse {
+    scanId: string;
+    code: WebApiErrorCodeName;
+    error: string;
+}
```
devaradhanm: we may need to send {
url,
code, 
error
}
& not scan id if there is error

packages/web-api/src/controllers/batch-scan-result-controller.ts
lamaks: Handled as part of ApiController

packages/web-api/src/api-contracts/scan-result-response.ts
lamaks: The ScanRunResponse is type when we have error before we assign scan id. Other cases will have scan id already so we can use above type.

PR #343, opened 2019-10-01
Added scan run error code placeholder. Unified all error response types.

lamaks:
#### Description of changes

Added scan run error code placeholder.
Unified all error response types.
Updated swagger schema.

#### Pull request checklist

### DIFF
--- a/packages/api-contracts/ai-scan-api.yaml
+++ b/packages/api-contracts/ai-scan-api.yaml
@@ -278,6 +278,7 @@ components:
                     type: string
                     enum:
                         - InternalError
+                        - UrlNavigationTimeout
                 message:
                     type: string
         ScanRunErrorResponse:
--- a/packages/api-contracts/ai-scan-api.yaml
+++ b/packages/api-contracts/ai-scan-api.yaml
@@ -251,6 +251,7 @@ components:
             type: object
             required:
                 - code
+                - codeId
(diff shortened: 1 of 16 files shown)


packages/api-contracts/ai-scan-api.yaml
```diff
@@ -237,17 +239,54 @@ components:
...
+                message:
+                    type: string
+        ScanRunErrorResponse:
+            type: object
+            required:
+                - scanId
+                - error
+            properties:
+                scanId:
+                    type: string
```
devaradhanm: shouldnt this be just url? since a job is not created in this case?

devaradhanm: there are two cases:
1.) you submit a scan jb on invalid url. for this error response could be:
{url: string, error: WebApiError}
2.) you ask for the scan status on invalid guid. for this response could be:
{scanId: string, error: WebApiError}

lamaks: No, this is scan error, not web api error. Any scan run (failed/pass) will have scan id
1. Invalid URL will not have scan id. We reject it at the injection time. If URL is not accessible from a browser this is scan run error that is returned in run object of scan run result
2. Invalid guid is handled on HTTP request and will return web api error before any processing

PR #351, opened 2019-10-02
Pownkel/dev installation fix

pownkel:
Fixed a bug in dev environment where, if a client id was given, it would still try to access service principal and add a reply URL. Fixed by checking if a client id was passed instead of only checking if profile is dev.


packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -212,7 +212,7 @@ deployWebApiFunctionApp() {
     deployWebApiArmTemplate $packageName
 
     # Add reply url to function app registration
-    if [ $environment = "dev" ]; then
+    if [ "$clientIdPassed" = false]; then
```
lamaks: Do we need double quotes around var here?

pownkel: Yes, if the variable is not defined (which it isn't unless no client id is given) then it would fail without the quotes

PR #356, opened 2019-10-04
Pownkel/vmss resource group retry

pownkel:
Verify that vmss resource group exists and can be accessed before trying to enable managed identity in installation script. If it fails, retry until a timeout.

### DIFF
--- a/packages/resource-deployment/scripts/batch-pool-enable-msi.sh
+++ b/packages/resource-deployment/scripts/batch-pool-enable-msi.sh
@@ -80,7 +80,8 @@ assignSystemIdentity() {
 
         # Wait until we are certain the resource group exists
         waiting=false
-        end=$((SECONDS + 600))
+        timeout=7200
+        end=$((SECONDS + $timeout))
         resourceGroupExists=$(az group exists --name "$vmssResourceGroup")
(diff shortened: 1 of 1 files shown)


packages/resource-deployment/scripts/batch-pool-enable-msi.sh
```diff
@@ -77,6 +77,32 @@ getVmssInfo() {
 assignSystemIdentity() {
     for vmssResourceGroup in "${vmssResourceGroups[@]}"; do
         echo "Enabling system-assigned managed identity for VMSS resource group $vmssResourceGroup"
+
+        # Wait until we are certain the resource group exists
+        waiting=false
+        timeout=1800 # timeout after half an hour
```
lamaks: Is this a case to wait for 30 mins?

pownkel: Yes, last week it was taking between 15 minutes and several hours for the resource group to become visible. There's a support ticket out right now for why it's taking so long, but we can reduce the timeout once that gets sorted out.

lamaks: Please keep related bug open to have the time reduced when Azure will have resolution.

PR #361, opened 2019-10-08
Pownkel/observe url response

### LINKED ISSUE #360: [Bug] Add error code for unreachable web page
If the url is unreachable, the reponse should have proper error code instead of returning internal server error.

pownkel:
Check the http status code when accessing a url for scanning. If it gives an error code, return this as a message to the user. (Fixes #360)


packages/scanner/src/page.ts
```diff
@@ -41,6 +41,10 @@ export class Page {
             return { unscannable: true, error: `Cannot scan ${url} because it is not a html page.` };
         }
```
devaradhanm: can you also take care of the exceptions that will be thrown by the goto url promise?
https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagegotourl-options

packages/scanner/src/page.spec.ts
```diff
@@ -8,6 +8,7 @@ import { AxePuppeteer } from 'axe-puppeteer';
 import * as Puppeteer from 'puppeteer';
 import { IMock, It, Mock, Times } from 'typemoq';
 
+import { ok } from 'assert';
```
devaradhanm: unused import

packages/scanner/src/page.ts
pownkel: Any exceptions thrown in scanning are caught by the Scanner class and passed on. Since we're now passing all error messages back to the user, exceptions from goto will already be returned to the user as well.

devaradhanm: so, we already have corresponding error codes for the exceptions returned by gotopromise?

packages/scanner/src/page.ts
```diff
@@ -41,6 +41,10 @@ export class Page {
...
 
+        if (!response.ok()) {
+            return { error: `Error accessing ${url}: ${response.status()} ${response.statusText()}` };
```
lamaks: Can we change the message similar to:
The URL request returned unsuccessful response ...

packages/web-api/src/converters/scan-run-error-converter.ts
```diff
@@ -10,6 +10,16 @@ export class ScanRunErrorConverter {
             return ScanRunErrorCodes.urlNavigationTimeout;
         }
 
-        return ScanRunErrorCodes.internalError;
+        return this.createInternalErrorWithMessage(scanError);
+    }
+
+    private createInternalErrorWithMessage(appendMessage: string): ScanRunErrorCode {
```
lamaks: We do not want to send our internal error message to a client as is. We may need to create a new error code for the Non Success HTTP response and match our message to be translated to a client message. RegEx can take portion w/ HTTP code from a message. Ideally the HTTP response code should be part of separate property in error response type.

packages/scanner/src/page.ts
pownkel: Added new error codes just now for the goto errors

packages/service-library/src/web-api/scan-run-error-codes.ts
```diff
@@ -26,4 +34,40 @@ export class ScanRunErrorCodes {
...
+    public static sslError: ScanRunErrorCode = {
+        code: 'SslError',
+        codeId: 9004,
+        message: 'SSL Error when navigating to scan URL',
+    };
+
+    public static mainResourceLoadFailure: ScanRunErrorCode = {
+        code: 'MainResourceLoadFailure',
+        codeId: 9005,
+        message: 'The main resource failed to load when navigating to scan URL',
```
lamaks: This is probably corresponding to "The resource is not available" message/code

packages/service-library/src/web-api/scan-run-error-codes.ts
```diff
@@ -26,4 +34,40 @@ export class ScanRunErrorCodes {
...
+    public static httpErrorCode: ScanRunErrorCode = {
+        code: 'HttpErrorCode',
+        codeId: 9003,
+        message: 'Accessing the scan URL resulted in an error.',
+    };
+
+    public static sslError: ScanRunErrorCode = {
+        code: 'SslError',
+        codeId: 9004,
+        message: 'SSL Error when navigating to scan URL',
```
lamaks: Neat: SSL error ...

PR #363, opened 2019-10-10
Pownkel/return redirected url

pownkel:
If the page to be scanned redirects, return the new redirected URL in a new field in the result.


packages/scanner/src/page.ts
```diff
@@ -60,7 +60,14 @@ export class Page {
         const axePuppeteer: AxePuppeteer = await this.axePuppeteerFactory.createAxePuppeteer(this.puppeteerPage);
         const scanResults = await axePuppeteer.analyze();
 
-        return { results: scanResults };
+        if (response.request().redirectChain().length > 0) {
```
flyingUnderTheRadar: General question : Is there any value in finding out the type of redirect? Some redirects are temporary, some are permanent. Does the WCP team care about this? 
Additionally is the WCP team interested in the number of redirects ? (I think they shouldn't care)

packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -97,6 +97,9 @@ export class Runner {
             pageScanResult.run = this.createRunResult('completed');
             pageScanResult.scanResult = this.getScanStatus(axeScanResults);
             pageScanResult.reports = [await this.saveScanReport(axeScanResults)];
+            if (axeScanResults.redirectedToUrl !== undefined) {
```
flyingUnderTheRadar: If the default value of `pageScanResult.redirectedToUrl` is `undefined` is this check necessary? Or are we trying to preserve an earlier set value here?

pownkel: Yes, if no value is set then that field doesn't get written

packages/scanner/src/page.spec.ts
```diff
@@ -216,7 +216,7 @@ describe('Page', () => {
 
         await page.enableBypassCSP();
 
-        const results = await page.scanForA11yIssues('https://www.bing.com');
+        const results = await page.scanForA11yIssues('https://tinyurl.com/mrhd5x');
```
haonliu: why do we have to change the url in this test case?

pownkel: My mistake, I was testing with tinyurl and forgot to change it back. That's fixed now.

packages/web-api/src/api-contracts/scan-result-response.ts
```diff
@@ -12,6 +12,7 @@ export declare type ScanResultResponse = ScanRunResultResponse | ScanRunErrorRes
 export interface ScanRunResultResponse {
     scanId: string;
     url: string;
+    redirectedToUrl?: string;
```
lamaks: Can we rename to scannedUrl ?

PR #367, opened 2019-10-17
Pownkel/health api

pownkel:
Added api endpoint that always returns 200 so we can ping and test if the server is running

### DIFF
--- a/packages/web-api/src/controllers/health-check-controller.ts
+++ b/packages/web-api/src/controllers/health-check-controller.ts
@@ -1,0 +1,31 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { ServiceConfiguration } from 'common';
+import { inject, injectable } from 'inversify';
+import { Logger } from 'logger';
+import { ApiController } from 'service-library';
+
+@injectable()
+export class HealthCheckController extends ApiController {
+    public readonly apiVersion = '1.0';
+    public readonly apiName = 'web-api-get-report';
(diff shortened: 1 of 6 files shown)


packages/web-api/src/controllers/health-check-controller.ts
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT License.
+import { ServiceConfiguration } from 'common';
+import { inject, injectable } from 'inversify';
+import { Logger } from 'logger';
+import { ApiController } from 'service-library';
+
+@injectable()
+export class HealthCheckController extends ApiController {
+    public readonly apiVersion = '1.0';
+    public readonly apiName = 'web-api-get-report';
```
lamaks: api name is incorrect

packages/web-api/src/controllers/health-check-controller.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    public constructor(
+        @inject(ServiceConfiguration) protected readonly serviceConfig: ServiceConfiguration,
+        @inject(Logger) private readonly logger: Logger,
+    ) {
+        super();
+    }
+
+    public async handleRequest(): Promise<void> {
+        this.logger.trackEvent('Health check success');
```
lamaks: trackEvent() accepts name of the event, not a message. We need to have App Insight events defined as a separate type that should be accepted by trackEvent() instead of an arbitrary string. Please work with @devaradhanm on implementation details.

PR #369, opened 2019-10-18
Added health monitoring functions placeholders

lamaks:
#### Description of changes

Added health monitoring functions placeholders.

#### Pull request checklist

### DIFF
--- a/packages/web-workers/health-monitor-client-func/function.json
+++ b/packages/web-workers/health-monitor-client-func/function.json
@@ -1,0 +1,10 @@
+{
+    "bindings": [
+        {
+            "name": "action",
+            "type": "activityTrigger",
+            "direction": "in"
+        }
+    ],
+    "scriptFile": "../health-monitor-client-func/index.js"
+}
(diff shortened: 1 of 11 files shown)


devaradhanm:
Can you explain the responsibility of each funtion api you are adding in this PR?

lamaks:
> Can you explain the responsibility of each funtion api you are adding in this PR?

The trigger function will runs in 5 minutes interval and start orchestration function. The orchestration function will call activity (client) function. The client function will call service api. Orchestration will have workflow to call rest api in a sequence with required delay. Orchestration function has restrictions on I/O operation that require to have client (activity) function.

devaradhanm:
Can you also add why we need orchestration & activity function instead of doing the work in time trigger itself? what are the limitations of using time trigger other than that the execution time affects the next schedule time.

lamaks:
> Can you also add why we need orchestration & activity function instead of doing the work in time trigger itself? what are the limitations of using time trigger other than that the execution time affects the next schedule time.

We need to persists scan id. Durable function can be handy here.
Orchestration I/O restriction https://docs.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-orchestrations#calling-http-endpoints

PR #370, opened 2019-10-21
Track Scan request batch submit event via app insights

haonliu:
#### Description of changes

image

#### Pull request checklist

### DIFF
--- a/packages/logger/src/logger-event-measurements.ts
+++ b/packages/logger/src/logger-event-measurements.ts
@@ -15,16 +15,14 @@ export interface BatchScanRequestMeasurements extends BaseTelemetryMeasurements 
 export interface BatchPoolMeasurements extends BaseTelemetryMeasurements {
     runningTasks: number;
     samplingIntervalInSeconds: number;
-    maxParallelTasksPossible: number;
+    maxParallelTasks: number;
 }
 
 export interface ScanTaskStartedMeasurements extends BaseTelemetryMeasurements {
     scanWaitTime: number;
-    scanTaskStartTime: number;
 }
 
 export interface ScanTaskCompletedMeasurements extends BaseTelemetryMeasurements {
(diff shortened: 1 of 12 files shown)


packages/logger/src/logger-event-measurements.ts
```diff
@@ -0,0 +1,30 @@
...
+    maxParallelTasksPossible: number;
+}
+
+export interface ScanTaskStartedMeasurements extends BaseTelemetryMeasurements {
+    scanWaitTime: number;
+    scanTaskStartTime: number;
+}
+
+export interface ScanTaskCompletedMeasurements extends BaseTelemetryMeasurements {
+    scanCompleteTime: number;
```
lamaks: scanCompletedTime
scanExecutionTime
scanWallClockTIme

packages/logger/src/logger-event-measurements.ts
```diff
@@ -0,0 +1,30 @@
...
+export interface BatchScanRequestMeasurements extends BaseTelemetryMeasurements {
+    totalScanRequests: number;
+    acceptedScanRequests: number;
+    rejectedScanRequests: number;
+}
+
+export interface BatchPoolMeasurements extends BaseTelemetryMeasurements {
+    runningTasks: number;
+    samplingIntervalInSeconds: number;
+    maxParallelTasksPossible: number;
```
lamaks: this is max actual running task, right?
should be just maxParallelTasks?

packages/logger/src/logger-event-measurements.ts
haonliu: fixed

packages/logger/src/logger-event-measurements.ts
haonliu: fixed

PR #377, opened 2019-10-23
Pownkel/correlation

pownkel:
Added ability to set common logger properties at runtime instead of passing them to each call, and used this to add scanId and batchRequestId to logger calls in web api runner and ScanRequestController


packages/logger/src/app-insights-logger-client.spec.ts
```diff
@@ -199,6 +201,97 @@ describe(AppInsightsLoggerClient, () => {
...
+            testSubject.trackException(error);
+            verifyMocks();
+        });
+
+        it('set custom dimensions for trackEvent', async () => {
+            setupCallsForTelemetrySetup();
+            await testSubject.setup(null);
+            appInsightsTelemetryClientMock.reset();
+
+            const customDimensionAssignments = { scanId: 'scan-id', batchRequestId: 'batch-req-id' };
```
haonliu: NIT: Can we add a test case in which one of the preset custom properties is overridden by the data passed into ``trackEvent()``?

packages/logger/src/app-insights-logger-client.ts
```diff
@@ -3,15 +3,19 @@
 import * as appInsights from 'applicationinsights';
 import { inject, injectable } from 'inversify';
 
+import { get } from 'http';
```
haonliu: Is this being used?

packages/logger/src/logger-properties.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+export interface LoggerProperties {
+    scanId?: string;
+    batchRequestId?: string;
```
haonliu: Should ``batchRequestId`` be required? Aren't we always going to have it?

pownkel: There won't be one if we decide to log anything before reading the request documents or assigning an id or something like that. Would it be better to not have the field in that case, or have the field exist but be empty?

packages/runner/src/runner/runner.ts
```diff
@@ -30,6 +30,7 @@ export class Runner {
         let browser: Browser;
         const runTime = new Date();
         const scanMetadata = this.scanMetadataConfig.getConfig();
+        this.logger.setCustomProperties({ scanId: scanMetadata.scanUrl });
```
haonliu: I'm not sure if we need to do this for the old flow.

PR #379, opened 2019-10-25
Pownkel/throughput metrics

pownkel:
Add InProgressScanRequests metric to track request throughput. Graphing the sum of this metric will give the total number of scan requests being processed (both in the queue and running scans) at a given time:

Throughput chart

### DIFF
--- a/packages/logger/src/app-insights-logger-client.spec.ts
+++ b/packages/logger/src/app-insights-logger-client.spec.ts
@@ -81,10 +81,10 @@ describe(AppInsightsLoggerClient, () => {
             appInsightsTelemetryClientMock.reset();
 
             appInsightsTelemetryClientMock
-                .setup(t => t.trackMetric(It.isValue({ name: 'metric1', value: 10, properties: {} })))
+                .setup(t => t.trackMetric(It.isValue({ name: 'InProgressScanRequests', value: 10, properties: {} })))
                 .verifiable();
 
-            testSubject.trackMetric('metric1', 10);
+            testSubject.trackMetric('InProgressScanRequests', 10);
 
             verifyMocks();
         });
@@ -238,7 +238,7 @@ describe(AppInsightsLoggerClient, () => {
                 .setup(t =>
                     t.trackMetric(
                         It.isValue({
-                            name: 'metric1',
+                            name: 'InProgressScanRequests',
                             value: 10,
                             properties: customDimensionAssignments,
                         }),
@@ -246,7 +246,7 @@ describe(AppInsightsLoggerClient, () => {
                 )
                 .verifiable();
             testSubject.setCustomProperties(customDimensionAssignments);
-            testSubject.trackMetric('metric1', 10);
+            testSubject.trackMetric('InProgressScanRequests', 10);
             verifyMocks();
         });
 
--- a/packages/logger/src/app-insights-logger-client.spec.ts
+++ b/packages/logger/src/app-insights-logger-client.spec.ts
@@ -81,10 +81,10 @@ describe(AppInsightsLoggerClient, () => {
             appInsightsTelemetryClientMock.reset();
 
             appInsightsTelemetryClientMock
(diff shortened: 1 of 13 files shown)


packages/web-api/src/controllers/scan-request-controller.ts
```diff
@@ -73,27 +74,38 @@ export class ScanRequestController extends ApiController {
...
                     url: scanRunRequest.url,
                 });
 
                 scanResponses.push({
                     scanId: scanId,
                     url: scanRunRequest.url,
                 });
+
+                if (requestCountByPriority[priority.toString()] === undefined) {
+                    requestCountByPriority[priority.toString()] = 0;
```
lamaks: Should we count first one ie set to 1

packages/logger/src/app-insights-logger-client.spec.ts
```diff
@@ -75,16 +75,30 @@ describe(AppInsightsLoggerClient, () => {
...
+        });
+
+        it('when properties passed', async () => {
             setupCallsForTelemetrySetup();
             await testSubject.setup(null);
             appInsightsTelemetryClientMock.reset();
 
             appInsightsTelemetryClientMock
-                .setup(t => t.trackMetric(It.isValue({ name: 'metric1', value: 10, properties: {} })))
+                .setup(t => t.trackMetric(It.isValue({ name: 'QueuedScanRequests', value: 1, properties: { foo: 'bar' } })))
```
haonliu: NIT: extract value and properties to a variable could help if there are changes to the method in the future.

packages/web-api/src/controllers/scan-request-controller.ts
```diff
@@ -73,27 +74,38 @@ export class ScanRequestController extends ApiController {
...
         const scanResponses: ScanRunResponse[] = [];
+        const requestCountByPriority: { [priority: string]: number } = {};
 
         scanRunRequests.forEach(scanRunRequest => {
             const runRequestValidationResult = this.validateRunRequest(scanRunRequest);
             if (runRequestValidationResult.valid) {
                 // preserve GUID origin for a single batch scope
                 const scanId = this.guidGenerator.createGuidFromBaseGuid(batchId);
+                const priority = isNil(scanRunRequest.priority) ? 0 : scanRunRequest.priority;
                 scanRequestsToBeStoredInDb.push({
```
haonliu: If 0 is the default priority, I think we should extract a variable like ``DefaultPriority`` and save it at a proper place.

lamaks: Default is applied when property is not set. Priority is int with range -1000..1000. Just set int default

packages/web-api/src/controllers/scan-request-controller.ts
haonliu: I'm wondering if we can use [splitting ](https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-charts#apply-filters-to-charts) feature so that we can split the data by priority in app insights instead of doing it here.

packages/web-api/src/controllers/scan-request-controller.spec.ts
```diff
@@ -158,6 +158,7 @@ describe(ScanRequestController, () => {
 
             // tslint:disable-next-line: no-null-keyword
             loggerMock.setup(lm => lm.trackEvent('BatchScanRequestSubmitted', null, expectedMeasurements)).verifiable();
+            loggerMock.setup(lm => lm.trackMetric('QueuedScanRequests', 1, { priority: '1' })).verifiable();
```
devaradhanm: we need to discuss about passing priority as property for this metric.

devaradhanm: i see some concerns with this approach. if we go with this approach, 
if we queue 3 p "0" urls & 2 p "-10" urls, then we shou7ld raise the following metrics:
QueuedScanRequests - 3 (P 0)
QueuedScanRequests - 2 (P -10)

Given that we support -1000 to 1000 as priority value, this is not feasible.

PR #383, opened 2019-11-01
Create rest api client library

haonliu:
#### Description of changes

- Create rest api client library.
- Moved Rest API Request and Response typings to service library package 
- Authentication is not included in this PR.

#### Pull request checklist


packages/web-api-client-library/src/a11y-service-client.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private getRequestOptions(requestBody?: Object): request.RequestPromiseOptions {
+        return {
+            json: requestBody,
+            qs: {
+                'api-version': this.apiVersion,
+            },
+            headers: {
+                'Content-Type': 'application/json',
```
devaradhanm: you might also need to pass auth token that you get by calling aad

packages/web-api-client-library/src/a11y-service-client.ts
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { injectable } from 'inversify';
+import * as request from 'request-promise';
+
+@injectable()
+export class A11yServiceClient {
+    constructor(private readonly baseUrl: string, private readonly apiVersion = '1.0', private readonly httpRequest = request) {}
```
devaradhanm: how do you make sure that you don't create more than one http agent?

.vscode/workspace.code-workspace
```diff
@@ -64,6 +64,10 @@
             "name": "web-workers",
             "path": "../packages/web-workers"
         },
+        {
+            "name": "web-api-client-library",
```
devaradhanm: may be just call it web-api-client

haonliu: updated.

packages/web-api-client-library/src/a11y-service-client.ts
haonliu: As discussed offline auth will not be included in this PR.

packages/web-api-client-library/src/a11y-service-client.ts
haonliu: We are using node js request package which provides option to use [forever-agent](https://github.com/request/forever-agent).

devaradhanm: does it create a new forever agent for every request, or does it reuse the same agent?

haonliu: discussed offline

haonliu: Verified that the client will reuse the same https agent.

packages/web-api-client/package.json
```diff
@@ -0,0 +1,44 @@
+{
+    "name": "web-api-client-library",
```
devaradhanm: call it web-api-client

packages/web-api-client/src/register-web-api-client-library-to-container.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import * as inversify from 'inversify';
+import { A11yServiceClient } from './a11y-service-client';
+
+export function registerWebApiClientLibraryToContainer(container: inversify.Container): inversify.Container {
```
devaradhanm: this should not be in this package. let this be inside the web-worker's container setup.

packages/web-api-client/package.json
haonliu: updated

PR #386, opened 2019-11-05
Add authentication to rest api client

haonliu:
#### Description of changes
To call our rest endpoints:
````
    const cred = new A11yServiceCredential('client-id', 'client-sec', 'resource-id', 'authority-url');
    const client = new A11yServiceClient(cred, 'https://apim-foo.net');

    client.postScanUrl('https://foo.com', 5);

````

#### Pull request checklist

### CODE REVIEW

File: packages/web-api-client/src/a11y-service-authentication-handler.ts [hunk 1]
```diff
@@ -0,0 +1,45 @@
...
+                    } else {
+                        resolve(tokenResponse as TokenResponse);
+                    }
+                },
+            );
+        });
+    }
+
+    public async getAuthHeaders(): Promise<RequestPromiseOptions> {
+        const accessToken = await this.getToken();
```
devaradhanm: check if the sdk supports caching the token. if not then u need to cache token & get token again on expiry.

packages/web-api-client/src/a11y-service-client.ts
```diff
@@ -1,8 +1,10 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 import { injectable } from 'inversify';
+import { merge } from 'lodash';
 import * as request from 'request-promise';
 import { ScanResultResponse, ScanRunRequest, ScanRunResponse } from 'service-library';
+import { A11yServiceAuthenticationHandler } from './a11y-service-authentication-handler';
 
 @injectable()
```
devaradhanm: this doesnt have to be injectable.

packages/web-api-client/src/a11y-service-authentication-handler.ts
haonliu: They are caching tokens: see line 136 here: https://github.com/AzureAD/azure-activedirectory-library-for-nodejs/blob/889b2babaca7fe192f7c0c651526a4adc53c38ea/lib/token-request.js

packages/web-api-client/src/a11y-service-client.spec.ts
```diff
@@ -6,34 +6,38 @@ import * as request from 'request-promise';
...
+            post: postMock.object,
+        };
+        credMock = Mock.ofType<A11yServiceCredential>(null);
 
-    afterEach(() => {
-        requestMock.verifyAll();
+        credMock
+            .setup(cm => cm.signRequest(requestStub))
+            .returns(async () => Promise.resolve(requestStub))
+            .verifiable();
```
devaradhanm: move this signRequest call to helper method & then call it in every test. (setupVerifiableSignRequestCall). this way it is clear that this call is expected in each test.

packages/web-api-client/src/a11y-service-client.spec.ts
```diff
@@ -42,35 +46,34 @@ describe(A11yServiceClient, () => {
...
     it('getScanStatus', async () => {
-        const scanId = 'scanid';
+        const scanId = 'scanId';
         const response = { statusCode: 200 };
-        requestMock
-            .setup(req => req.get(`${baseUrl}/scans/${scanId}`))
-            // tslint:disable-next-line: no-any
-            .returns(() => Promise.resolve(response) as any)
+
+        getMock
```
devaradhanm: add 1 more test when priority is not set. (make sure all the branches are covered by uts)

packages/web-api-client/src/a11y-service-credential.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+        requestMock = Mock.ofType<typeof request>(null);
+        authenticationContextMock = Mock.ofType<AuthenticationContext>();
+
+        testSubject = new A11yServiceCredential(clientId, clientSecret, resource, authorityUrl, authenticationContextMock.object);
+
+        authenticationContextMock
+            .setup(am => am.acquireTokenWithClientCredentials(resource, clientId, clientSecret, It.isAny()))
+            .returns((resourceUrl, cid, sec, callback) => {
+                callback(error, tokenResponse);
+            });
```
devaradhanm: move this to a helper method & then call it in each test method. (setupVerifiableGetTokenCall)

haonliu: Will quickly send a PR to address.

PR #395, opened 2019-11-08
add app reg sec names in typing

haonliu:
#### Description of changes

add app reg sec names in typing

#### Pull request checklist


packages/azure-services/src/key-vault/secret-names.ts
```diff
@@ -5,4 +5,6 @@ export const secretNames = {
     cosmosDbKey: 'cosmosDbKey',
     storageAccountName: 'storageAccountName',
     storageAccountKey: 'storageAccountKey',
+    serverAppRegistrationSecret: 'A11yServiceServerAppRegistration',
```
manishsat: remind me what these are used for?

haonliu: They are used in authentication. In web worker we only need serverAppRegistrationSecret, in the future like e2e test we will use clientAppRegistrationSecret (consumer app).

PR #404, opened 2019-11-13
Adding Availability test by calling all our rest apis

devaradhanm:
#### Description of changes
With this PR we have added changes required to call the rest apis using activity trigger & then use app insights' trackAvaliability api to measure availability
Still unit tests need to be added

#### Pull request checklist

### DIFF
--- a/packages/common/src/configuration/service-configuration.ts
+++ b/packages/common/src/configuration/service-configuration.ts
@@ -35,6 +35,7 @@ export interface ScanRunTimeConfig {
 export interface RestApiConfig {
     maxScanRequestBatchCount: number;
     scanRequestProcessingDelayInSeconds: number;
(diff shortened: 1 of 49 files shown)


packages/common/src/configuration/service-configuration.ts
```diff
@@ -35,6 +35,7 @@ export interface ScanRunTimeConfig {
 export interface RestApiConfig {
     maxScanRequestBatchCount: number;
     scanRequestProcessingDelayInSeconds: number;
+    maxScanRequestWaitTimeInSeconds: number;
```
haonliu: NIT: This does not look like a config for rest api.

devaradhanm: it is not. we can move it out later.

PR #414, opened 2019-11-18
#405 - Single req support

### LINKED ISSUE #405: [Feature] Remove batching from our API's
## Is your feature request related to a problem? Please describe.

Spoke with @manishsat and he mentioned batching is no longer required and we should remove batching from our API spec and implementations.

## Describe the solution you'd like

This work will include changing the API spec as well as cutting down unwanted code in the API layer only. Both submit and get will no longer be batched.
- Post scans will take in a single URL with priority
- There will be a single GET scan endpoint that takes in a single url. The batch endpoint will be dropped as part of a different feature.
- The GET report endpoint remains unchanged.

Scope is limited to changing the API spec and API layer. Everything else will remain as is and everything underneath is still handled the same way. A separate feature will work to remove the flab and re-architect everything below the API layer.

Work with Uday and come up with a way to transition smoothly.

flyingUnderTheRadar:
#### Description of changes

This adds a v2.0 that makes the req / response for the POST scans to accept only a single url and respond with a single object.

#### Pull request checklist

- [X] Addresses an existing issue: Fixes #405
- [X] Added relevant unit test for your changes. (`yarn test`)
- [X] Ran precheckin (`yarn precheckin`)

### DIFF
--- a/packages/service-library/src/web-api/api-controller.ts
+++ b/packages/service-library/src/web-api/api-controller.ts
@@ -74,7 +74,7 @@ export abstract class ApiController extends WebController {
             return false;
         }
 
-        if (this.context.req.query['api-version'] !== this.apiVersion) {
+        if (this.context.req.query['api-version'] !== this.apiVersion && this.context.req.query['api-version'] !== '2.0') {
             this.context.res = HttpResponse.getErrorResponse(WebApiErrorCodes.unsupportedApiVersion);
 
             return false;
--- a/packages/service-library/src/web-api/web-api-error-codes.ts
+++ b/packages/service-library/src/web-api/web-api-error-codes.ts
@@ -127,7 +127,7 @@ export class WebApiErrorCodes {
         error: {
             code: 'MalformedBody',
             codeId: 4011,
-            message: 'Please validate the request body against the schema for this version and endpoint.',
+            message: 'The request body does not match the API schema for the given API version.',
         },
     };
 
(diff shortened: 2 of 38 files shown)


packages/service-library/src/web-api/web-api-error-codes.ts
```diff
@@ -121,6 +122,15 @@ export class WebApiErrorCodes {
         },
     };
 
+    public static malformedRequest: WebApiErrorCode = {
+        statusCode: 400,
+        error: {
+            code: 'MalformedBody',
+            codeId: 4011,
+            message: 'Please validate the request body against the schema for this version and endpoint.',
```
lamaks: May be rephrase to something like: The request body does not match to the API schema for the given API version.

packages/service-library/src/web-api/api-controller.ts
```diff
@@ -74,7 +74,7 @@ export abstract class ApiController extends WebController {
             return false;
         }
 
-        if (this.context.req.query['api-version'] !== this.apiVersion) {
+        if (this.context.req.query['api-version'] !== this.apiVersion && this.context.req.query['api-version'] !== '2.0') {
```
lamaks: We can change this.apiVersion to array since the single controller implementation supports two versions

flyingUnderTheRadar: Synced offline - This is okay for now.

packages/service-library/src/web-api/web-api-error-codes.ts
flyingUnderTheRadar: Done.

PR #427, opened 2019-11-26
Pownkel/generate html report

pownkel:
#### Description of changes

Consume html generation package to generate scan report

#### Pull request checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,6 +146,6 @@
         "https-proxy-agent": "^3.0.1"
     },
     "dependencies": {
-        "markreay-accessibility-insights-report": "0.0.1"
+        "markreay-accessibility-insights-report": "0.0.3"
     }
 }
--- a/package.json
+++ b/package.json
@@ -144,8 +144,5 @@
         "axe-core": "^3.3.2",
         "lodash": "^4.17.14",
         "https-proxy-agent": "^3.0.1"
-    },
-    "dependencies": {
-        "accessibility-insights-report": "0.1.0"
     }
 }
--- a/package.json
+++ b/package.json
@@ -144,5 +144,8 @@
         "axe-core": "^3.3.2",
         "lodash": "^4.17.14",
         "https-proxy-agent": "^3.0.1"
+    },
+    "dependencies": {
+        "markreay-accessibility-insights-report": "0.0.1"
     }
 }
--- a/package.json
+++ b/package.json
@@ -146,6 +146,6 @@
         "https-proxy-agent": "^3.0.1"
     },
     "dependencies": {
-        "markreay-accessibility-insights-report": "0.0.3"
+        "accessibility-insights-report": "0.1.0"
     }
 }
--- a/packages/web-api-scan-runner/package.json
+++ b/packages/web-api-scan-runner/package.json
@@ -65,6 +65,6 @@
         "storage-documents": "1.0.0",
         "verror": "^1.10.0",
         "yargs": "^13.2.1",
-        "markreay-accessibility-insights-report": "^0.0.1"
+        "markreay-accessibility-insights-report": "^0.0.3"
     }
 }
--- a/packages/web-api-scan-runner/package.json
+++ b/packages/web-api-scan-runner/package.json
@@ -65,6 +65,6 @@
         "storage-documents": "1.0.0",
         "verror": "^1.10.0",
         "yargs": "^13.2.1",
-        "markreay-accessibility-insights-report": "^0.0.3"
+        "accessibility-insights-report": "^0.1.0"
     }
 }
--- a/packages/web-api-scan-runner/package.json
+++ b/packages/web-api-scan-runner/package.json
@@ -64,6 +64,7 @@
         "sha.js": "^2.4.11",
         "storage-documents": "1.0.0",
         "verror": "^1.10.0",
-        "yargs": "^13.2.1"
+        "yargs": "^13.2.1",
+        "markreay-accessibility-insights-report": "^0.0.1"
     }
 }
--- a/packages/web-api-scan-runner/src/report-generator/axe-result-converter.ts
+++ b/packages/web-api-scan-runner/src/report-generator/axe-result-converter.ts
@@ -10,7 +10,7 @@ export type ReportGenerationParams = {
 };
 
 export interface AxeResultConverter {
-    readonly reportType: ReportFormat;
+    readonly targetReportFormat: ReportFormat;
 
     convert(results: AxeResults, params: ReportGenerationParams): string;
 }
--- a/packages/web-api-scan-runner/src/report-generator/axe-result-converter.ts
+++ b/packages/web-api-scan-runner/src/report-generator/axe-result-converter.ts
@@ -2,16 +2,14 @@
 // Licensed under the MIT License.
 
 import { AxeResults } from 'axe-core';
-import { injectable } from 'inversify';
 import { ReportFormat } from 'storage-documents';
 
 export type ReportGenerationParams = {
     pageTitle: string;
 };
 
-@injectable()
-export abstract class AxeResultConverter {
-    public readonly reportType: ReportFormat;
+export interface AxeResultConverter {
+    readonly reportType: ReportFormat;
 
-    public abstract convert(results: AxeResults, params: ReportGenerationParams): string;
+    convert(results: AxeResults, params: ReportGenerationParams): string;
 }
--- a/packages/web-api-scan-runner/src/report-generator/axe-result-converter.ts
+++ b/packages/web-api-scan-runner/src/report-generator/axe-result-converter.ts
@@ -1,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { AxeResults } from 'axe-core';
+import { injectable } from 'inversify';
+import { ReportFormat } from 'storage-documents';
+
+export type ReportGenerationParams = {
+    pageTitle: string;
+};
+
+@injectable()
+export abstract class AxeResultConverter {
+    public readonly reportType: ReportFormat;
+
+    public abstract convert(results: AxeResults, params: ReportGenerationParams): string;
+}
--- a/packages/web-api-scan-runner/src/report-generator/axe-result-converter.ts
+++ b/packages/web-api-scan-runner/src/report-generator/axe-result-converter.ts
@@ -6,6 +6,7 @@ import { ReportFormat } from 'storage-documents';
 
 export type ReportGenerationParams = {
     pageTitle: string;
+    browserSpec: string;
 };
 
 export interface AxeResultConverter {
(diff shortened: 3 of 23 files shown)


packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -24,13 +25,24 @@ import { WebDriverTask } from '../tasks/web-driver-task';
 
 // tslint:disable: no-null-keyword no-any
 
-function convertAxeToHtml(results: AxeResults): string {
-    return 'html report';
+function convertAxeToHtml(results: AxeResults, pageTitle: string): string {
```
devaradhanm: let this be a class AxeResultToHtmlConverter & then let the class take reporterFactory via constructor & use dependency injection.

devaradhanm:
```
@injectable()
export class AxeResultToHtmlConverter {
    constructor(@inject(iocTypes.ReporterFactory) private readonly repoertFactory: typeof reporterFactory) {}
    convert(results: AxeResults, pageTitle: string): string {}
}
```

devaradhanm: another refactoring could be, you could have a ReportGenerator class that is responsible of creating the two report types. & will return the report[] ({reportType: 'html'| 'sarif', reportId: string})

package.json
```diff
@@ -144,5 +144,8 @@
         "axe-core": "^3.3.2",
         "lodash": "^4.17.14",
         "https-proxy-agent": "^3.0.1"
+    },
+    "dependencies": {
+        "markreay-accessibility-insights-report": "0.0.3"
```
lamaks: This was updated to public version

packages/web-api-scan-runner/package.json
```diff
@@ -64,6 +64,7 @@
         "sha.js": "^2.4.11",
         "storage-documents": "1.0.0",
         "verror": "^1.10.0",
-        "yargs": "^13.2.1"
+        "yargs": "^13.2.1",
+        "markreay-accessibility-insights-report": "^0.0.3"
```
lamaks: Do we need to have this in both root and package specific level? Should it be only here? Need to be public version as well.

pownkel: Looks like it's fine with it only being in the package level. I've removed it from the root.

packages/web-api-scan-runner/src/report-generator/axe-result-converter.ts
```diff
@@ -0,0 +1,17 @@
...
+import { AxeResults } from 'axe-core';
+import { injectable } from 'inversify';
+import { ReportFormat } from 'storage-documents';
+
+export type ReportGenerationParams = {
+    pageTitle: string;
+};
+
+@injectable()
+export abstract class AxeResultConverter {
```
devaradhanm: this should be interface, if it doesnt have any implementation

packages/web-api-scan-runner/src/report-generator/report-generator.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+        reportGenerator = new ReportGenerator(guidGeneratorMock.object, axeResultConverters);
+    });
+
+    it('calls convert on all axeResultConverters', () => {
+        reportGenerator.generateReports(axeResults);
+
+        // tslint:disable-next-line:prefer-const
+        for (let axeResultConverter of axeResultConverters) {
+            expect(axeResultConverter.convertCallCount).toBe(1);
+        }
```
devaradhanm: NIT: you could use foreach.
```
axeResultConverters.forEach(axeResultConverter=> {
 expect(axeResultConverter.convertCallCount).toBe(1);
});

```

packages/web-api-scan-runner/src/report-generator/axe-result-to-html-converter.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    public convert(results: AxeResults, params: ReportGenerationParams): string {
+        const reporter = this.reporterFactoryFunc();
+
+        const htmlReportParams: AxeReportParameters = {
+            results: results,
+            description: 'Automated report',
+            serviceName: 'Accessibility Insights Service',
+            scanContext: {
+                browserSpec: 'BROWSER_SPEC',
```
haonliu: Could we pass some meaningful string as the browser spec?

packages/web-api-scan-runner/src/report-generator/axe-result-to-html-converter.ts
```diff
@@ -0,0 +1,34 @@
...
+    constructor(@inject(iocTypeNames.ReporterFactory) private readonly reporterFactoryFunc: ReporterFactory) {
+        super();
+    }
+
+    public convert(results: AxeResults, params: ReportGenerationParams): string {
+        const reporter = this.reporterFactoryFunc();
+
+        const htmlReportParams: AxeReportParameters = {
+            results: results,
+            description: 'Automated report',
```
lamaks: Can it be more descriptive? What is PM recommendation here?

packages/web-api-scan-runner/src/report-generator/axe-result-to-sarif-converter.ts
```diff
@@ -0,0 +1,24 @@
...
+import { AxeResults } from 'axe-core';
+import { SarifLog } from 'axe-sarif-converter';
+import { inject, injectable } from 'inversify';
+import { ReportFormat } from 'storage-documents';
+import { iocTypeNames } from '../ioc-types';
+import { AxeResultConverter, ReportGenerationParams } from './axe-result-converter';
+
+@injectable()
+export class AxeResultToSarifConverter extends AxeResultConverter {
+    public readonly reportType: ReportFormat = 'sarif';
```
lamaks: This is not used here. Can we have public type of report formats declared outside so it can be referenced in a solution?

packages/web-api-scan-runner/src/report-generator/axe-result-to-html-converter.ts
```diff
@@ -0,0 +1,34 @@
...
+import { AxeReportParameters, ReporterFactory } from 'accessibility-insights-report';
+import { AxeResults } from 'axe-core';
+import { inject, injectable } from 'inversify';
+import { ReportFormat } from 'storage-documents';
+import { iocTypeNames } from '../ioc-types';
+import { AxeResultConverter, ReportGenerationParams } from './axe-result-converter';
+
+@injectable()
+export class AxeResultToHtmlConverter extends AxeResultConverter {
+    public readonly reportType: ReportFormat = 'html';
```
lamaks: This is not used here. Can we have public type of report formats declared outside so it can be referenced in a solution?

packages/web-api-scan-runner/src/report-generator/axe-result-to-html-converter.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    constructor(@inject(iocTypeNames.ReporterFactory) private readonly reporterFactoryFunc: ReporterFactory) {}
+
+    public convert(results: AxeResults, params: ReportGenerationParams): string {
+        const reporter = this.reporterFactoryFunc();
+
+        const htmlReportParams: AxeReportParameters = {
+            results: results,
+            description: 'Automated report',
+            serviceName: 'Accessibility Insights Service',
```
haonliu: NIT: these strings could be put at a central place like what we are doing in web repo: https://github.com/microsoft/accessibility-insights-web/blob/master/src/content/strings/application.ts

packages/web-api-scan-runner/src/report-generator/report-generator.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { GuidGenerator } from 'common';
+import { inject, injectable } from 'inversify';
+import { AxeScanResults } from 'scanner';
+import { ReportFormat } from 'storage-documents';
+import { iocTypeNames } from '../ioc-types';
+import { AxeResultConverter } from './axe-result-converter';
+
+export type GeneratedReport = {
+    report: string;
```
lamaks: May be rename to contents or body?

packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -107,10 +93,7 @@ export class Runner {
             this.logger.trackEvent('ScanTaskSucceeded');
             pageScanResult.run = this.createRunResult('completed');
             pageScanResult.scanResult = this.getScanStatus(axeScanResults);
-            pageScanResult.reports = [
-                await this.saveScanReport(axeScanResults, 'sarif'),
-                await this.saveScanReport(axeScanResults, 'html'),
-            ];
+            pageScanResult.reports = await Promise.all(this.generateAndSaveScanReports(axeScanResults));
```
lamaks: Better move Promise.all() inside to the call map() as it bit misleading that promise.all() should be used here

packages/web-api-scan-runner/src/setup-web-api-scan-runner-container.ts
```diff
@@ -15,5 +20,14 @@ export function setupWebApiScanRequestSenderContainer(): inversify.Container {
     registerScannerToContainer(container);
     registerServiceLibraryToContainer(container);
 
+    container.bind(iocTypeNames.ConvertAxeToSarifFunc).toConstantValue(convertAxeToSarif);
```
lamaks: This can be part of the scanner package like other register*() belongs to corresponding packages

packages/web-api-scan-runner/src/report-generator/axe-result-to-sarif-converter.ts
pownkel: resolved via offline discussion

packages/web-api-scan-runner/src/report-generator/axe-result-to-html-converter.ts
pownkel: resolved via offline discussion

packages/web-api-scan-runner/src/report-generator/axe-result-converter.ts
```diff
@@ -0,0 +1,16 @@
...
+import { AxeResults } from 'axe-core';
+import { ReportFormat } from 'storage-documents';
+
+export type ReportGenerationParams = {
+    pageTitle: string;
+    browserSpec: string;
+};
+
+export interface AxeResultConverter {
+    readonly reportType: ReportFormat;
```
lamaks: May be rename to resultReportFormat or targetReportFormat to indicate the converter target format operation

packages/web-api-scan-runner/src/report-generator/axe-result-to-html-converter.ts
pownkel: Addie suggested adding scan url and time. I've updated this to include those.

PR #438, opened 2019-12-16
Consistent logger use

### LINKED ISSUE #432: [Bug] Inconsistent HTTP request singleton pattern
The solution had single logging provider Logger() in a corresponding library /Logger. To support HTTP request singleton pattern the new logging provider was introduced ContextAwareLogger(). The sparse usage of a new logging provider introduced inconsistent logging pattern where some packages use Logger() (for example the azure-services package) and web-api package uses ContextAwareLogger().

As result the Logger() indirectly being used in HTTP context where ContextAwareLogger() is used as well which do not let log context related event properly.

Expected implementation pattern
---------------------------------
The context aware pattern should be implemented on IoC container level instead. Any other providers that require context awareness should be part on that container (like logging library). The context IoC container should follow HTTP request singleton pattern.

pownkel:
#### Description of changes

Instead of binding both Logger and ContextAwareLogger in the IoC container, bind one object to Logger and use either a GlobalLogger or ContextAwareLogger as appropriate for the context. Fixes #432 
- Split registerLoggerToContainer into registerGlobalLoggerToContainer and registerContextAwareLoggerToContainer
- Remove exports of logger implementations (GlobalLogger and ContextAwareLogger) and only export the base Logger class
- bind a ContextAwareLogger in request-level containers
- Add a few additional setup steps in contextAwareLoggerClient (make sure rootLoggerClient is setup)

#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #432 

### DIFF
--- a/packages/logger/src/base-app-insights-logger-client.ts
+++ b/packages/logger/src/base-app-insights-logger-client.ts
@@ -13,9 +13,13 @@ import { LoggerEvent } from './logger-event';
 export abstract class BaseAppInsightsLoggerClient implements LoggerClient {
     protected telemetryClient: TelemetryClient;
 
+    protected initialized: boolean = false;
+
     public abstract async setup(baseProperties?: BaseTelemetryProperties): Promise<void>;
 
-    public abstract isSetup(): boolean;
+    public isInitialized(): boolean {
+        return this.initialized;
+    }
 
     public trackMetric(name: string, value: number): void {
         this.telemetryClient.trackMetric({
--- a/packages/logger/src/base-app-insights-logger-client.ts
+++ b/packages/logger/src/base-app-insights-logger-client.ts
@@ -15,6 +15,8 @@ export abstract class BaseAppInsightsLoggerClient implements LoggerClient {
 
     public abstract async setup(baseProperties?: BaseTelemetryProperties): Promise<void>;
 
+    public abstract isSetup(): boolean;
+
     public trackMetric(name: string, value: number): void {
         this.telemetryClient.trackMetric({
             name: name,
--- a/packages/logger/src/context-aware-app-insights-logger-client.ts
+++ b/packages/logger/src/context-aware-app-insights-logger-client.ts
@@ -20,13 +20,14 @@ export class ContextAwareAppInsightsLoggerClient extends BaseAppInsightsLoggerCl
             ...baseProperties,
         };
 
-        if (!this.rootLoggerClient.isSetup()) {
+        if (!this.rootLoggerClient.isInitialized()) {
             await this.rootLoggerClient.setup();
         }
+        this.initialized = true;
     }
 
-    public isSetup(): boolean {
-        return this.telemetryClient !== undefined && this.rootLoggerClient.isSetup();
+    public isInitialized(): boolean {
+        return super.isInitialized() && this.rootLoggerClient.isInitialized();
     }
 
     protected getAdditionalPropertiesToAddToEvent(): { [key: string]: string } {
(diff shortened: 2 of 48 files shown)


packages/logger/src/context-aware-app-insights-logger-client.ts
```diff
@@ -18,6 +19,14 @@ export class ContextAwareAppInsightsLoggerClient extends BaseAppInsightsLoggerCl
...
             ...baseProperties,
         };
+
+        if (!this.rootLoggerClient.isSetup()) {
+            await this.rootLoggerClient.setup();
+        }
+    }
+
+    public isSetup(): boolean {
+        return this.telemetryClient !== undefined && this.rootLoggerClient.isSetup();
```
lamaks: Moving `private initialized: boolean` in a base class as `protected` and have only `public isSetup()` in a base class allow simplify implementation in inherited class by directly setting/getting `initialized` value directly

packages/logger/src/base-app-insights-logger-client.ts
```diff
@@ -15,6 +15,8 @@ export abstract class BaseAppInsightsLoggerClient implements LoggerClient {
 
     public abstract async setup(baseProperties?: BaseTelemetryProperties): Promise<void>;
 
+    public abstract isSetup(): boolean;
```
lamaks: IsInitialized() is more common name

packages/logger/src/context-aware-logger.ts
```diff
@@ -5,7 +5,6 @@ import { injectable } from 'inversify';
 import { Logger } from './logger';
 import { LoggerClient } from './logger-client';
 
-@injectable()
 export class ContextAwareLogger extends Logger {
```
lamaks: Remove import { injectable }...

packages/logger/src/global-logger.ts
```diff
@@ -6,7 +6,6 @@ import { BaseTelemetryProperties } from './base-telemetry-properties';
 import { Logger } from './logger';
 import { LoggerClient } from './logger-client';
 
-@injectable()
```
lamaks: Remove import { injectable }...

packages/logger/src/register-logger-to-container.ts
```diff
@@ -38,11 +25,34 @@ export function registerLoggerToContainer(container: Container): void {
...
+    container.bind(loggerTypes.AppInsights).toConstantValue(appInsights);
+    container.bind(loggerTypes.Process).toConstantValue(process);
+    container
+        .bind(AppInsightsLoggerClient)
+        .toSelf()
+        .inSingletonScope();
+    container
+        .bind(ConsoleLoggerClient)
+        .toSelf()
+        .inSingletonScope();
```
lamaks: This is context related bind for `registerGlobalLoggerToContainer()`

pownkel: Since it uses inSingletonScope there will be one global object for AppInsightsLoggerClient and ConsoleLoggerClient

packages/web-workers/src/process-web-request.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    beforeEach(() => {
+        requestHandled = false;
+        context = ({
+            req: {
+                query: { 'api-version': '1.0' },
+            },
+        } as unknown) as Context;
+        process.env.APPINSIGHTS_INSTRUMENTATIONKEY = '00000000-0000-0000-0000-000000000000';
+        //console.log = jest.fn();
```
haonliu: Do we need this?

pownkel: Nope, I thought I took that out, my mistake

packages/logger/src/register-logger-to-container.ts
lamaks: Right, but this is not a part of  `registerLoggerDependenciesToContainer()` and not need to be called from `registerContextAwareLoggerToContainer()`. The `registerLoggerDependenciesToContainer()` is intended to be common for both registrations

PR #442, opened 2019-12-20
Report test runs in health api

pownkel:
#### Description of changes

Health API should query Application Insights for test results and return a report of which tests are passing or failing. In this PR:
- Change health endpoint API to return an empty test report
- Inject ApplicationInsightsClient into health check controller and execute a placeholder query
- Add process of generating and saving an app insights api key to deployment scripts

#### Pull request checklist


packages/web-api/src/controllers/health-check-controller.ts
```diff
@@ -13,14 +15,30 @@ export class HealthCheckController extends ApiController {
...
         @inject(ServiceConfiguration) protected readonly serviceConfig: ServiceConfiguration,
         @inject(Logger) logger: Logger,
+        @inject(webApiTypeNames.ApplicationInsightsClientProvider)
+        protected readonly appInsightsClientProvider: ApplicationInsightsClientProvider,
     ) {
         super(logger);
     }
 
     public async handleRequest(): Promise<void> {
         this.logger.trackEvent('HealthCheck');
```
haonliu: I'm a bit concerned about querying app insights in the health check end point. 
Feel free to correct me: if app insights is down, would the health check throw error? If so health check api will return 404 but our service is actually alive.

pownkel: Good point, I'll add a check to handle those errors

PR #444, opened 2020-01-01
Request sender queue size generator integration

### LINKED ISSUE #346: [Bug] Requests storage queue size should correlate with batch pool size
To support scan priority processing the request storage queue size should be small enough to not hold priority scan request for too long. Optimal queue size should be 2x of a single batch processing ie priority message should be red no longer than 2x time of average single message  processing.
To achieve this batch pool manager calculated size data need to be connected with scan request sender.

lamaks:
#### Description of changes

Request sender queue size generator integration

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #346


packages/azure-services/src/azure-batch/batch.ts
```diff
@@ -23,12 +24,21 @@ export class Batch {
...
-        @inject(iocTypeNames.BatchTaskParameterProvider) private readonly batchTaskParameterProvider: BatchTaskParameterProvider,
+        @optional()
+        @inject(iocTypeNames.BatchTaskParameterProvider)
+        private readonly batchTaskParameterProvider: BatchTaskParameterProvider,
         @inject(StorageContainerSASUrlProvider) private readonly containerSASUrlProvider: StorageContainerSASUrlProvider,
         @inject(BatchConfig) private readonly config: BatchConfig,
         @inject(Logger) private readonly logger: Logger,
     ) {}
 
+    public async getJobScheduleRunIntervalInMinutes(): Promise<number> {
```
devaradhanm: ut missing

packages/web-api-scan-job-manager/src/worker/worker.ts
```diff
@@ -112,13 +111,24 @@ export class Worker {
         return jobQueuedTasks.length;
     }
 
+    private async getPoolLoadSnapshot(poolMetricsInfo: PoolMetricsInfo): Promise<PoolLoadSnapshot> {
+        const lastPoolLoadSnapshot = await this.systemDataProvider.readBatchPoolLoadSnapshot(this.batchConfig.accountName, 'urlScanPool');
+        const poolLoadSnapshot = await this.poolLoadGenerator.getPoolLoadSnapshot(lastPoolLoadSnapshot.activityState, poolMetricsInfo);
```
devaradhanm: can lastPoolLoadSnapshot  be null or undefined here?

packages/azure-services/src/azure-batch/pool-load-generator.ts
```diff
@@ -95,10 +117,20 @@ export class PoolLoadGenerator {
...
     }
+
+    // tslint:disable: no-bitwise
+    private getActivityState(lastActivityState: number, isIdle: boolean): number {
+        const mask = 2147483647;
+        // shift previous states left and reset sign bit
+        const nextState = (lastActivityState << 1) & mask;
+
+        // set current state bit to 0/1 for idle/active pool respectively
+        return isIdle ? nextState : nextState + 1;
```
devaradhanm: why do we need this? can't we just deal with state as strings? like 'idle', 'running' etc?

packages/azure-services/src/system-config.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { injectable } from 'inversify';
+
+@injectable()
+export class SystemConfig {
+    public readonly storageName: string = process.env.AZURE_STORAGE_NAME;
```
devaradhanm: it might be better to not use AZURE_STORAGE_NAME as this is set only after QueueServiceURLProvider or BlobServiceClientProvider is used. Also, caching secrets in general can be solved by adding a caching layer for all the fetched secrets.

packages/common/src/configuration/service-configuration.ts
```diff
@@ -79,16 +79,20 @@ export class ServiceConfiguration {
...
                         config.loadFile(ServiceConfiguration.profilePath);
                         config.validate({ allowed: 'strict' });
-                    } else {
-                        console.log(`Unable to load custom configuration. Using default config  - ${config}`);
+                        configExists = true;
                     }
-                    resolve(config);
                 });
+
+                if (!configExists) {
```
devaradhanm: this code should be inside the fs.exists callback as before

packages/service-library/src/data-providers/system-data-provider.ts
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under the MIT License.
+import { CosmosContainerClient, cosmosContainerClientTypes } from 'azure-services';
+import { inject, injectable } from 'inversify';
+import { BatchPoolLoadSnapshot, ItemType, PartitionKey, ScanQueueLoadSnapshot } from 'storage-documents';
+
+export declare type BatchPoolAlias = 'urlScanPool';
+export declare type StorageQueueAlias = 'onDemandScanRequest';
+
+@injectable()
+export class SystemDataProvider {
```
devaradhanm: rename it to be more specific. may be TaskProcessingStatisticsDataProvider?

packages/web-api-scan-job-manager/src/setup-web-api-scan-job-manager-container.ts
```diff
@@ -19,11 +19,9 @@ export function setupWebApiScanJobManagerContainer(): Container {
         .inSingletonScope();
 
     container
-        .bind(ScanTaskParameterProvider)
-        .toSelf()
+        .bind<BatchTaskParameterProvider>(AzureServicesIocTypes.BatchTaskParameterProvider)
+        .to(ScanTaskParameterProvider)
```
devaradhanm: can u update the ut in setup-web-api-scan-job-manager-container.spec.ts to also validate singleton resolution for this type under "resolves service config to singleton value" test?

packages/azure-services/src/system-config.ts
lamaks: used in conjunction with queue storage hence it will be initialized

packages/service-library/src/data-providers/system-data-provider.ts
lamaks: ScanProcessingStateProvider

packages/azure-services/src/azure-batch/pool-load-generator.ts
lamaks: this is continuous pool states represented as bits

Issue #430, opened 2019-12-03
[Bug] Gracefully handle batch job manager restart


lamaks:
Currently the batch job manager terminates even there are corresponding scan tasks pending/running.

The expected job manager termination workflow:
1. Stop queuing batch task
2. Wait for pending task to complete
3. Terminate job manager

PR #447, opened 2020-01-03
waiting for child tasks to complete before terminating job manager

devaradhanm:
#### Description of changes
Addresses issue #430

#### Pull request checklist

### DIFF
--- a/packages/job-manager/src/worker/worker.ts
+++ b/packages/job-manager/src/worker/worker.ts
@@ -39,7 +39,7 @@ export class Worker {
                 const scanMessages = await this.getMessages(tasksIncrementCount);
                 if (scanMessages.length === 0) {
                     this.logger.logInfo(`The storage queue '${this.queue.scanQueue}' has no message to process.`);
-                    if (poolMetricsInfo.load.activeTasks === 0 && poolMetricsInfo.load.runningTasks === 1) {
+                    if (this.hasChildTasksRunning(poolMetricsInfo) === false) {
                         this.logger.logInfo(`Exiting the ${this.jobId} job since there are no active/running tasks.`);
                         break;
                     }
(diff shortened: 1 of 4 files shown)


packages/job-manager/src/worker/worker.ts
```diff
@@ -65,12 +66,24 @@ export class Worker {
...
+            await this.system.wait(this.jobManagerConfig.addTasksIntervalInSeconds * 1000);
+        }
+    }
+
+    private async waitForChildTasks(): Promise<void> {
+        this.logger.logInfo('Waiting for child tasks to complete');
+
+        let poolMetricsInfo: PoolMetricsInfo = await this.batch.getPoolMetricsInfo();
+
+        while (poolMetricsInfo.load.activeTasks !== 0 || poolMetricsInfo.load.runningTasks > 1) {
```
manishsat: Don't you think the termination condition should be 
if (poolMetricsInfo.load.activeTasks === 0 && poolMetricsInfo.load.runningTasks === 1) as per line 42?

devaradhanm: both are same. (negation of what i have) in this case, i keep waiting if there are any active task or running task

packages/web-api-scan-job-manager/src/worker/worker.ts
```diff
@@ -73,12 +74,24 @@ export class Worker {
...
+            await this.system.wait(this.jobManagerConfig.addTasksIntervalInSeconds * 1000);
+        }
+    }
+
+    private async waitForChildTasks(): Promise<void> {
+        this.logger.logInfo('Waiting for child tasks to complete');
+
+        let poolMetricsInfo: PoolMetricsInfo = await this.batch.getPoolMetricsInfo();
+
+        while (poolMetricsInfo.load.activeTasks !== 0 || poolMetricsInfo.load.runningTasks > 1) {
```
lamaks: you can generalize the exit case and use it from above: 

The Batch service API may set activeTasks value instead of runningTasks value hence handle this case
--
  | if (poolMetricsInfo.load.runningTasks + poolMetricsInfo.load.activeTasks === 1) {
  | this.logger.logInfo(`Exiting the ${this.jobId} job since there are no active/running tasks.`);
  | break;
  | }

PR #450, opened 2020-01-06
Release versioning

pownkel:
#### Description of changes

Save release id as an env variable in function app so we can query a specific build's test results. New script (update-release-version.sh) should be added as a new step in the build pipeline after restarting batch pools so that we don't accidentally query test runs on a system that is not fully updated.

#### Pull request checklist


packages/resource-deployment/scripts/update-release-version.sh
```diff
@@ -0,0 +1,51 @@
...
+getFunctionAppNames() {
+    functionAppNamesString=$(az functionapp list --resource-group "$resourceGroupName" --query '[].name' -o tsv | tr "\n" ",")
+    echo "Found function apps: $functionAppNamesString"
+    IFS=$',' read -ra functionAppNames <<<"$functionAppNamesString"
+}
+
+setReleaseVersion() {
+    functionAppName=$1
+    echo "Updating release version for function app $functionAppName"
+    az functionapp config appsettings set --name $functionAppName --resource-group $resourceGroupName --settings "RELEASE_VERSION=$releaseVersion" 1>/dev/null
```
devaradhanm: i would suggest to do this as part of the arm template. this way, it is consistent with how we set other env variables.

devaradhanm: (like in function-app-template.json)

packages/web-api/src/controllers/health-check-controller.ts
```diff
@@ -48,7 +54,7 @@ export class HealthCheckController extends ApiController {
         };
     }
 
-    // Override this method not to check api version
+    // Override thqgitis method not to check api version
```
lamaks: typo

packages/web-api/src/controllers/health-check-controller.ts
```diff
@@ -26,20 +26,26 @@ export class HealthCheckController extends ApiController {
...
             testsFailed: 0,
         };
 
-        const e2eTestConfig = await this.serviceConfig.getConfigValue('e2eTestConfig');
-        const queryString = 'customEvents | limit 5';
-        const queryResponse = await appInsightsClient.executeQuery(queryString, e2eTestConfig.testRunQueryTimespan);
-        if (queryResponse.statusCode === 200) {
-            this.logger.logInfo(`App insights api queried with result ${JSON.stringify(queryResponse)}`);
+        if (currentVersion === undefined) {
+            healthReport.error = 'No release version number is configured';
```
lamaks: we probably should follow same approach we have with rest api and return HTTP 400 Bad Request with message body as we have in other apis

pownkel: Do we want to do that if it's a configuration issue with the function app, not a problem with the request?

packages/resource-deployment/scripts/update-release-version.sh
haonliu: Are we going to do it inside the arm template?

packages/web-api/src/controllers/health-check-controller.ts
lamaks: if we have problem with configuration we should send HTTP 500 instead

PR #468, opened 2020-01-15
Add script to evaluate functional test results

haonliu:
#### Description of changes

Add script to evaluate functional test results.
Also made change to the service client so that we can get the health report.

#### Pull request checklist


packages/functional-tests/src/validate-e2e.ts
```diff
@@ -0,0 +1,52 @@
...
+
+let healthStatus: TestRunResult;
+
+setTimeout(async () => {
+    while (healthStatus !== 'pass') {
+        try {
+            console.log('[E2E] fetching health report');
+            const response = await client.checkHealth(`/release/${argv.releaseId}`);
+            if (response.statusCode !== 200) {
+                throw new Error(response.statusMessage);
```
lamaks: we can print entire response since status message may not be sufficient

packages/functional-tests/src/validate-e2e.ts
```diff
@@ -0,0 +1,52 @@
...
+            }
+            healthStatus = response.body.healthStatus;
+        } catch (error) {
+            console.log('[E2E] failed to evaluate functional test results: ', error);
+        }
+
+        if (healthStatus !== 'pass') {
+            setTimeout(() => {
```
lamaks: we need to log the health response details to see in a pipeline what is falling

packages/functional-tests/src/validate-e2e.ts
```diff
@@ -0,0 +1,52 @@
...
+const cred = new A11yServiceCredential(argv.clientId, argv.clientSecret, argv.clientId, argv.authorityUrl);
+const client = new A11yServiceClient(cred, argv.baseUrl);
+
+const waitTimeBeforeEvaluation = parseInt(argv.waitTimeBeforeEvaluationInMinutes) * 60 * 1000;
+const evaluationInterval = parseInt(argv.evaluationIntervalInMinutes) * 60 * 1000;
+
+let healthStatus: TestRunResult;
+
+setTimeout(async () => {
+    while (healthStatus !== 'pass') {
```
lamaks: we may need validation task timeout as well to terminate wait

haonliu: there is a timeout option in azure devops task config. We can use that.

packages/functional-tests/src/validate-e2e.ts
haonliu: good point. printing it regardless whether it is failing or passing.

packages/functional-tests/src/validate-e2e.ts
```diff
@@ -0,0 +1,53 @@
...
+
+        if (healthStatus !== 'pass') {
+            setTimeout(() => {
+                console.log('[E2E] waiting for next evaluation');
+            }, evaluationInterval);
+        }
+    }
+}, waitTimeBeforeEvaluation);
+
+console.log('[E2E] functional tests passed.');
```
devaradhanm: this line shouldnt be after the settimeout. Curently, this line will be printed even before the settimeout callback is executed.

packages/functional-tests/src/validate-e2e.ts
devaradhanm: you should still have this fail after trying n times so that it fails gracefully instead of just throwing a timeout error.

packages/functional-tests/src/validate-e2e.ts
```diff
@@ -0,0 +1,54 @@
...
+console.log('[E2E] - start evaluate functional test results');
+
+const cred = new A11yServiceCredential(argv.clientId, argv.clientSecret, argv.clientId, argv.authorityUrl);
+const client = new A11yServiceClient(cred, argv.baseUrl);
+
+const waitTimeBeforeEvaluation = parseInt(argv.waitTimeBeforeEvaluationInMinutes) * 60 * 1000;
+const evaluationInterval = parseInt(argv.evaluationIntervalInMinutes) * 60 * 1000;
+
+let healthStatus: TestRunResult;
+
```
devaradhanm: add a comment before settimeout that you are waiting

packages/functional-tests/src/validate-e2e.ts
lamaks: we are using a group task and timeout is hidden within group, better to have it explicitly in a code

haonliu: updated.

packages/functional-tests/src/validate-e2e.ts
```diff
@@ -0,0 +1,69 @@
...
+            healthStatus = response.body.healthStatus;
+        } catch (error) {
+            console.log('[E2E] failed to evaluate functional test results: ', error);
+        }
+
+        if (healthStatus !== 'pass') {
+            if (isTestTimeOut(testStartTime, new Date(), testTimeoutInMinutes * 60 * 1000)) {
+                console.log('[E2E] e2e tests validation timeout');
+
+                throw new Error('[E2E] e2e tests validation timeout');
```
devaradhanm: if healthstatus is not pass, you should still print what tests have failed

haonliu: I'm printing the response regardless of the test results.

PR #469, opened 2020-01-15
Delete Batch pool load system data persistence

lamaks:
#### Description of changes

Remove unused code
Update pool load to normalize stat data returner based on pool idle/running state
Fix binding of the BatchTaskParameterProvider

#### Pull request checklist


packages/azure-services/src/azure-batch/batch.ts
```diff
@@ -117,25 +119,37 @@ export class Batch {
         };
     }
 
+    private async getJobSchedule(jobScheduleId: string): Promise<JobScheduleGetResponse> {
```
devaradhanm: this private method is not used anywhere

packages/common/src/configuration/service-configuration.ts
```diff
@@ -140,7 +140,7 @@ export class ServiceConfiguration {
                 },
                 maxWallClockTimeInHours: {
                     format: 'int',
```
devaradhanm: why 29?

lamaks: revert to initial implementation since batch api that was returning active/running task issue is fixed

packages/azure-services/src/azure-batch/pool-load-generator.spec.ts
```diff
@@ -45,16 +45,19 @@ describe(PoolLoadGenerator, () => {
             },
         };
         const expectedPoolLoadSnapshot = {
+            isIdle: false,
             tasksIncrementCountPerInterval: 60,
             targetActiveToRunningTasksRatio: 2,
             configuredMaxTasksPerPool: poolMetricsInfo.maxTasksPerPool,
+            targetMaxTasksPerPool: 64,
```
devaradhanm: if targetMaxTasksPerPool is being used for debugging purpose only ( to store in app insights), it is still misleading, why we even have this? we are logging the maxTasksPerPool config & tasksIncrementCountPerInterval(amount of tasks that can be added in next interval).

lamaks: tasksIncrementCountPerInterval is dynamic delta increment value where is targetMaxTasksPerPool is reflecting overall capacity/throughput of the pool currently running. There is no any issue/harm with having this values in result object as it is logged in app insight and may help do debugging/investigation

PR #483, opened 2020-01-18
Add option to run e2e tests locally

haonliu:
#### Description of changes

Add local e2e test script.

Provide the option to run functional test locally before check in. 

- if we make some changes to az function, instead of deploy the whole service and run the dev pipeline to check the functional test results, we can just publish the function and then run the e2e tests to verify the changes. 
- when we only changed the client side code, we don't need to deploy anything, just run the e2e tests locally.
- when we add more test cases into e2e tests, we can verify them easily.
- make it easier for debugging purpose.

#### Pull request checklist


packages/functional-tests/src/functional-tests.local.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+import { FunctionalTestGroup } from './test-groups/functional-test-group';
+
+// tslint:disable: mocha-no-side-effect-code no-any no-unsafe-any mocha-unneeded-done strict-boolean-expressions
+
+describe('functional tests', () => {
+    // need to provide by user. DO NOT CHECK IN
+    const clientId = process.env.SP_CLIENT_ID || '';
+    const clientSecret = process.env.SP_PASSWORD || '';
+    const tenantId = process.env.SP_TENANT || '';
+    const apimName = process.env.APIM_Name || '';
```
lamaks: APIM_SERVICE_NAME

packages/functional-tests/src/functional-tests.local.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+    let a11yServiceClient: A11yServiceClient;
+    let testRunner: TestRunner;
+    let guidGenerator: GuidGenerator;
+    let testContextData: TestContextData;
+    let cosmosClientWrapper: CosmosClientWrapper;
+    let cosmosContainerClient: CosmosContainerClient;
+    let onDemandPageScanRunResultProvider: OnDemandPageScanRunResultProvider;
+
+    beforeAll(async () => {
+        if (isServiceCredProvided()) {
```
lamaks: if not provided do we want to indicate that?

packages/functional-tests/src/functional-tests.local.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+            testRunner = new TestRunner(logger);
+            guidGenerator = new GuidGenerator();
+            cosmosClientWrapper = new CosmosClientWrapper(async () => {
+                return new CosmosClient({ endpoint: cosmosUrl, auth: { masterKey: cosmosKey } });
+            }, logger);
+            cosmosContainerClient = new CosmosContainerClient(cosmosClientWrapper, dbName, collectionName, logger);
+            onDemandPageScanRunResultProvider = new OnDemandPageScanRunResultProvider(
+                cosmosContainerClient,
+                new PartitionKeyFactory(new HashGenerator(), guidGenerator),
+            );
```
lamaks: It seems  most of this code already implemented when we setup DI container. Can we reuse DI container implementation?

packages/functional-tests/src/functional-tests.local.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+    const clientSecret = process.env.SP_PASSWORD || '';
+    const tenantId = process.env.SP_TENANT || '';
+    const apimName = process.env.APIM_Name || '';
+    const cosmosKey = process.env.COSMOS_DB_KEY || '';
+    const cosmosUrl = process.env.COSMOS_DB_URL || '';
+    // static values
+    const dbName = 'onDemandScanner';
+    const collectionName = 'scanRuns';
+    const releaseIdStub = 'release-id';
+    const runIdStub = 'run-id';
```
lamaks: Better to have real values here since it goes to app insight and will have to debug

haonliu: updated.

packages/functional-tests/src/functional-tests.local.spec.ts
haonliu: Updated.

haonliu: Reused some DI implementation that is available. Need to refactor the 'getProcessLifeCycleContainer' in web-worker before we can fully reuse that. will do that in a separate PR.

packages/functional-tests/src/functional-tests.local.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+        const scanReportsTests = getTests('ScanReports');
+        await testRunner.run(scanReportsTests, TestEnvironment.all, releaseIdStub, runIdStub);
+        done();
+    });
+
+    testIf('Finalizer', isServiceCredProvided, async (done: jest.DoneCallback) => {
+        const finalizerTests = getTests('Finalizer');
+        await testRunner.run(finalizerTests, TestEnvironment.all, releaseIdStub, runIdStub);
+        done();
+    });
```
lamaks: It looks very similar implementation. Can we do some generic function that get list of tests groups to run in a sequence?

packages/functional-tests/src/functional-tests.local.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+    testIf('ScanStatus', isServiceCredProvided, async (done: jest.DoneCallback) => {
+        const scanStatusTests = getTests('ScanStatus');
+        testContextData.scanId = (await a11yServiceClient.postScanUrl(testContextData.scanUrl)).body[0].scanId;
+        await testRunner.run(scanStatusTests, TestEnvironment.all, releaseIdStub, runIdStub);
+        done();
+    });
+
+    testIf('ScanExecution', isServiceCredProvided, async (done: jest.DoneCallback) => {
+        const scanRunResultResponse = await waitForScanRequestCompletion();
+        expect(scanRunResultResponse.reports.length).toEqual(2);
```
lamaks: Should this function just run e2e tests w/o any validation? If we need validation it should be part of an e2e test instead

packages/functional-tests/src/functional-tests.local.spec.ts
```diff
@@ -0,0 +1,159 @@
...
+
+import { TestEnvironment } from './common-types';
+import { functionalTestGroupTypes, TestGroupName } from './functional-test-group-types';
+import { TestRunner } from './runner/test-runner';
+import { TestContextData } from './test-group-data';
+import { FunctionalTestGroup } from './test-groups/functional-test-group';
+
+// tslint:disable: mocha-no-side-effect-code no-any no-unsafe-any mocha-unneeded-done strict-boolean-expressions
+
+describe('functional tests', () => {
```
lamaks: In general would be better to have this implemented as a generic e2e tests group runner similar to what we have in health orchestration worker

lamaks: Reuse the health REST API result can be part of this, so we can unify implementation and do not fork e2e runners

packages/functional-tests/src/functional-tests.local.spec.ts
haonliu: yep. updated.

packages/functional-tests/src/functional-tests.local.spec.ts
haonliu: removed.

packages/functional-tests/src/functional-tests.local.spec.ts
haonliu: Agreed. that could be in a future PR?

PR #488, opened 2020-01-24
Added handler to complete scan tasks when terminated unexpectedly

### LINKED ISSUE #431: [Bug] The batch scan task will not have scan DB state updated when terminated by Batch service
The Batch service will terminate long running scan tasks. In this case no corresponding state will be updated within Cosmom DB. The scan state will be marked as 'running'.

There should be a mechanism to handle batch scan tasks state after external termination. Either job manager should re-validate tasks states periodically or job scheduler post task should do re-validation.

lamaks:
#### Description of changes

Preserve task arguments object as part of task environment variable
Complete terminated tasks by setting correct scan state in Cosmos DB
Delete obsolete install script
Fix Mac OS bash compatibility issue

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #431, #491


packages/web-api-scan-job-manager/src/worker/worker.ts
```diff
@@ -76,18 +81,58 @@ export class Worker {
...
+                    let pageScanResult = await this.onDemandPageScanRunResultProvider.readScanRun(taskArguments.id);
+                    if (pageScanResult !== undefined) {
+                        if (pageScanResult.run.state !== 'failed') {
+                            pageScanResult.run = {
+                                state: 'failed',
+                                timestamp: failedTask.timestamp.toJSON(),
+                                error,
+                            };
+                            pageScanResult = await this.onDemandPageScanRunResultProvider.updateScanRun(pageScanResult);
+                        }
```
devaradhanm: can you log here the failure info from batch here? incase we have 2 different failures (scan failure message is different from task failure)

lamaks: batch error logged below
`                    this.logger.logError(error, { taskProperties: JSON.stringify(failedTask) });
`

packages/web-api-scan-job-manager/src/worker/worker.spec.ts
```diff
@@ -39,6 +38,7 @@ describe(Worker, () => {
     let serviceConfigMock: IMock<ServiceConfiguration>;
     let loggerMock: IMock<MockableLogger>;
     let batchPoolLoadSnapshotProviderMock: IMock<BatchPoolLoadSnapshotProvider>;
+    let onDemandPageScanRunResultProviderMock: IMock<OnDemandPageScanRunResultProvider>;
```
devaradhanm: i don't see onDemandPageScanRunResultProviderMock.verifyAl() call anywhere. Without this call, the verifiable setups won't be validated.

packages/azure-services/src/azure-batch/batch.spec.ts
```diff
@@ -98,6 +116,171 @@ describe(Batch, () => {
...
+            taskMock
+                .setup(async o => o.list(config.jobId, options))
+                .returns(async () => Promise.resolve(<TaskListResponse>(<unknown>items1)))
+                .verifiable();
+            taskMock
+                .setup(async o => o.listNext(items1.odatanextLink, options))
+                .returns(async () => Promise.resolve(<TaskListResponse>(<unknown>items2)))
+                .verifiable();
+
+            const expectedFailedTasks: BatchTask[] = [
```
devaradhanm: can you refactor this & create a helper method to generate these objects?

packages/web-api-scan-job-manager/src/worker/worker.spec.ts
lamaks: `afterEach(() => {
        batchMock.verifyAll();
        queueMock.verifyAll();
        serviceConfigMock.verifyAll();
        batchPoolLoadSnapshotProviderMock.verifyAll();
        systemMock.verifyAll();
        onDemandPageScanRunResultProviderMock.verifyAll();
    });`

packages/azure-services/src/azure-batch/batch.spec.ts
lamaks: No, generalization in UTs is not always the best and will not help future understanding of test data/test flow. In this case it has no high benefits either as only few properties are common in array of objects. The simple declaration is only required here as well.

PR #512, opened 2020-02-21
chore: Add optional parameter for chrome path

AhmedAbdoOrtiga:
#### Description of changes

Add optional parameter for chrome path so the action can pass it, it will be used only for the library.

#### Pull request checklist

### DIFF
--- a/packages/cli/src/scanner/page.ts
+++ b/packages/cli/src/scanner/page.ts
@@ -20,9 +20,9 @@ export class Page {
         @inject(WebDriver) private readonly webDriver: WebDriver,
     ) {}
 
-    public async create(): Promise<void> {
+    public async create(chromePath?: string): Promise<void> {
         if (isEmpty(this.browser)) {
-            this.browser = await this.webDriver.launch();
+            this.browser = await this.webDriver.launch(chromePath);
         }
         this.puppeteerPage = await this.browser.newPage();
     }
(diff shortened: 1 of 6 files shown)


packages/cli/src/scanner/page.ts
```diff
@@ -20,9 +20,9 @@ export class Page {
         @inject(WebDriver) private readonly webDriver: WebDriver,
     ) {}
 
-    public async create(): Promise<void> {
+    public async create(chromePath?: string): Promise<void> {
         if (isEmpty(this.browser)) {
-            this.browser = await this.webDriver.launch();
+            this.browser = await this.webDriver.launch(chromePath);
```
haonliu: I feel like chrome path should be a property of the web driver. Maybe we should pass it into the constructor of WebDriver?

AhmedAbdoOrtiga: No, we only wanna expose it to the library, and we don't want any breaking changes in the CLI

devaradhanm: @haonliu, we may need to refactor this to may be look something like this - https://github.com/devaradhanm/hello-world-javascript-action/blob/master/browser/web-driver.ts

devaradhanm: but we can do that later

packages/cli/src/web-driver/web-driver.ts
```diff
@@ -9,12 +9,19 @@ export class WebDriver {
...
+        let launchOption: Puppeteer.LaunchOptions;
+        launchOption = {
             headless: true,
             timeout: 30000,
             args: ['--disable-dev-shm-usage'],
-        });
+        };
+
+        if (chromePath !== undefined) {
+            launchOption.executablePath = chromePath;
```
devaradhanm: codecov says you don't have test for this

PR #580, opened 2020-03-10
chore: Adding recreate-pool script

devaradhanm:
#### Description of changes

- As part of this PR, I have refactored our deployment scripts & fixed some bugs;
          - Removed redundant subscription argument to make it easier to run scripts.
          - Passing -y argument when installing az func-core tools to avoid prompt with y/n causing script to abort since no user interaction is possible when running via script. 
          - Refactored to create set-resource-names.sh making it easier to pass less number of arguments when running scripts individually.
- Refactored to create script run-command-on-all-vmss-for-pool.sh & can run any command on all vmss in batch pools.
- Added a recreate pool script. We only run this script if the vmss are old by looking at vmssCreatedDate tag added in this PR(This disables all job schedule; wait for pool to be idle; resize to zero; scale up back by deploying batch template again; enabling all job schedules) (recreate-vmss-for-pools.sh )
- install-parallel will run the recreate-pools script always after this change.

Validated refactoring & recreate pool scripts here - https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_releaseProgress?_a=release-pipeline-progress&releaseId=3229

#### Pull request checklist


packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -52,10 +52,10 @@ installAzureFunctionsCoreToolsOnLinux() {
 
     # Verify your Ubuntu server is running one of the appropriate versions from the table below. To add the apt source, run
     sudo sh -c 'echo "deb [arch=amd64] https://packages.microsoft.com/repos/microsoft-ubuntu-$(lsb_release -cs)-prod $(lsb_release -cs) main" > /etc/apt/sources.list.d/dotnetdev.list'
-    sudo apt-get update
+    sudo apt-get update -y
 
     # Install the Core Tools package
-    sudo apt-get install azure-functions-core-tools=2.7.1948-1
+    sudo apt-get install azure-functions-core-tools=2.7.1948-1 -y
```
devaradhanm: passing -y to continue upgrade & avoid aborting script.

packages/resource-deployment/scripts/set-resource-names.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
lamaks: this is more `get` than `set` script

packages/resource-deployment/scripts/batch-account-create.sh
```diff
@@ -48,53 +90,16 @@ if ! az account show 1>/dev/null; then
...
-fi
-
-if [[ -z $enableSoftDeleteOnKeyVault ]]; then
-    enableSoftDeleteOnKeyVault=true
-fi
-
-# Login into Azure Batch account
-echo "Logging into '$batchAccountName' Azure Batch account"
-az batch account login --name "$batchAccountName" --resource-group "$resourceGroupName"
+. "${0%/*}/push-secrets-to-key-vault.sh"
```
lamaks: not logically belongs to this script, can we have it in top level script?

packages/resource-deployment/scripts/install-parallel.sh
```diff
@@ -153,7 +149,6 @@ runInParallel parallelProcesses
 # Additionally, these should run sequentially because of interdependence.
 
 . "${0%/*}/batch-account-create.sh"
-. "${0%/*}/push-secrets-to-key-vault.sh"
```
lamaks: echo to batch-account-create comment - can we keep it at top level since hide it in batch-account-create not intuitive

packages/resource-deployment/scripts/push-secrets-to-key-vault.sh
```diff
@@ -118,26 +118,25 @@ createAppInsightsApiKey() {
...
     *) exitWithUsageInfo ;;
     esac
 done
 
+
 # Print script usage help
-if [[ -z $cosmosAccountName ]] || [[ -z $resourceGroupName ]] || [[ -z $storageAccountName ]] || [[ -z $keyVault ]] || [[ -z $webApiAdClientId ]] || [[ -z $webApiAdClientSecret ]] || [[ -z $appInsightsName ]]; then
-    echo "$cosmosAccountName $resourceGroupName $storageAccountName $keyVault $appInsightsName"
+if [[ -z $resourceGroupName ]] || [[ -z $webApiAdClientId ]] || [[ -z $webApiAdClientSecret ]]; then
+    echo "$resourceGroupSuffix $resourceGroupName"
```
lamaks: not clear what this message for, should it have additional message text?

packages/resource-deployment/scripts/set-resource-names.sh
```diff
@@ -0,0 +1,29 @@
...
+fi
+
+storageAccountName=$(az storage account list \
+                        --resource-group "$resourceGroupName" \
+                        --query "[?starts_with(name,'allystorage')].name|[0]" \
+                        -o tsv
+                    )
+
+if [[ -z "$storageAccountName" ]]; then
+    echo "Unable to get storage account."
```
lamaks: ... for resource group `$resourceGroupName`

packages/resource-deployment/scripts/wait-for-deployment.sh
```diff
@@ -28,7 +28,7 @@ for arg in "$@"; do
     previousFlag=$arg
 done
 
-if [ -z $resourceName ] || [ -z $timeout ]; then
+if [ -z "$resourceName" ] || [ -z "$timeout" ]; then
```
lamaks: curious why the `""` are required only here? should this look like `if [[ ... ]]`

packages/resource-deployment/scripts/add-tags-for-batch-vmss.sh
```diff
@@ -0,0 +1,43 @@
...
+addResourceGroupNameTagToVMSS(){
+    az resource update --set tags.ResourceGroupName="$resourceGroupName" -g "$vmssResourceGroup" -n "$vmssName" --resource-type "Microsoft.Compute/virtualMachineScaleSets"
+
+    echo "Tag ResourceGroupName=$resourceGroupName was added to $vmssName vmss under $vmssResourceGroup resource group"
+}
+
+# Read script arguments
+while getopts ":v:r:b:" option; do
+    case $option in
+    v) vmssName=${OPTARG} ;;
```
AhmedAbdoOrtiga: can we generalize it more; 
- Send the resource type too as an input
- Rename the file to "add-tags-for-resource.sh"

packages/resource-deployment/scripts/add-tags-for-batch-vmss.sh
```diff
@@ -0,0 +1,43 @@
...
+
+exitWithUsageInfo() {
+    echo "
+Usage: $0 -v <vmss name> -r <vmss resource group> -b <batch resource group name>
+"
+    exit 1
+}
+
+addResourceGroupNameTagToVMSS(){
+    az resource update --set tags.ResourceGroupName="$resourceGroupName" -g "$vmssResourceGroup" -n "$vmssName" --resource-type "Microsoft.Compute/virtualMachineScaleSets"
```
AhmedAbdoOrtiga: can you call the new script you created for adding a tag instead of this line?

packages/resource-deployment/scripts/run-command-on-all-vmss-for-pool.sh
```diff
@@ -76,67 +80,54 @@ getVmssInfo() {
...
-        echo "Enabling system-assigned managed identity for VMSS resource group $vmssResourceGroup"
+        echo "Running setup for VMSS resource group $vmssResourceGroup for pool $pool"
 
         # Wait until we are certain the resource group exists
         . "${0%/*}/wait-for-deployment.sh" -n "$vmssResourceGroup" -t "1800" -q "az group exists --name $vmssResourceGroup"
 
         vmssQueryConditions="?tags.PoolName=='$pool' && tags.BatchAccountName=='$batchAccountName' && resourceGroup=='$vmssResourceGroup'"
         vmssDeployedQuery="[$vmssQueryConditions && provisioningState!='Creating' && provisioningState!='Updating'].name"
-        . "${0%/*}/wait-for-deployment.sh" -n "$vmssResourceGroup" -t "1800" -q "az vmss list --query \"$vmssDeployedQuery\" -o tsv"
+        vmssCreatedCommand="az vmss list --query \"$vmssDeployedQuery\" -o tsv"
```
lamaks: vmssCreatedFilter

packages/resource-deployment/scripts/run-command-on-all-vmss-for-pool.sh
```diff
@@ -76,67 +80,54 @@ getVmssInfo() {
...
+"
 # Read script arguments
-while getopts ":r:a:p:" option; do
+while getopts ":r:p:c:n:" option; do
     case $option in
     r) resourceGroupName=${OPTARG} ;;
-    a) batchAccountName=${OPTARG} ;;
     p) pool=${OPTARG} ;;
+    c) command=${OPTARG} ;;
+    n) commandName=${OPTARG} ;;
```
lamaks: probably this is redundant option, if we print command in echo above it may be more informative otherwise

packages/resource-deployment/scripts/run-command-on-all-vmss-for-pool.sh
```diff
@@ -76,67 +80,54 @@ getVmssInfo() {
 
 setupVmss() {
     for vmssResourceGroup in "${vmssResourceGroups[@]}"; do
-        echo "Enabling system-assigned managed identity for VMSS resource group $vmssResourceGroup"
+        echo "Running setup for VMSS resource group $vmssResourceGroup for pool $pool"
```
lamaks: the script comment needs to be updated `# The script will enable system-assigned managed identity on Batch pool VMSS`

packages/resource-deployment/scripts/run-command-on-all-vmss-for-pool.sh
```diff
@@ -76,67 +80,54 @@ getVmssInfo() {
...
-    vmssName=$2
-    
-    az resource update --set tags.ResourceGroupName="$resourceGroupName" -g "$vmssResourceGroup" -n "$vmssName" --resource-type "Microsoft.Compute/virtualMachineScaleSets"
-
-    echo "Tag ResourceGroupName=$resourceGroupName was added to $vmssName vmss under $vmssResourceGroup resource group"
-}
-
+. "${0%/*}/set-resource-names.sh"
+
+echo "inside pool setup"
```
lamaks: do you mean `Running pool setup for`?

packages/resource-deployment/scripts/batch-account-create.sh
devaradhanm: keyvault is created by batch. Also, for batch nodes to be up & running, we need keyvault to be setup.
having this here so that, nodes don't startup with failed tasks & run till timeout.

packages/resource-deployment/scripts/install-parallel.sh
devaradhanm: replied in the other comment.

packages/resource-deployment/scripts/add-tags-for-batch-vmss.sh
devaradhanm: we can generalize it further when we need it for other case. I am also doing some batch specific thing here.

packages/resource-deployment/scripts/add-tags-for-batch-vmss.sh
devaradhanm: this is the only file script i have for tags.

packages/resource-deployment/scripts/run-command-on-all-vmss-for-pool.sh
devaradhanm: command can have sensitive information. Hence, added command name

packages/resource-deployment/scripts/setup-all-pools-for-batch.sh
```diff
@@ -0,0 +1,84 @@
...
+}
+
+function setupPools() {
+    # Enable managed identity on Batch pools
+    pools=$(az batch pool list --query "[].id" -o tsv)
+
+    echo "Setup system identity for created pools"
+    for pool in $pools; do
+        command=". ${0%/*}/enable-system-identity-for-batch-vmss.sh"
+        commandName="System identity for pool $pool"
```
lamaks: not correlate with command - do we need the command name?

devaradhanm: as mentioned above, it can have sensitve data.

packages/resource-deployment/scripts/setup-all-pools-for-batch.sh
```diff
@@ -0,0 +1,84 @@
...
+        echo "Process with pid $pid exited"
+    done
+}
+
+function setupPools() {
+    # Enable managed identity on Batch pools
+    pools=$(az batch pool list --query "[].id" -o tsv)
+
+    echo "Setup system identity for created pools"
+    for pool in $pools; do
```
lamaks: can this run in parallel?

devaradhanm: nope. key vault access policy cannot be set in parallel for the same keyvault. It throws conflict error & ends up removing some access policy.

packages/resource-deployment/scripts/add-tags-for-batch-vmss.sh
AhmedAbdoOrtiga: yes that's the one I mean, that script for tags runs exactly the same command here

packages/resource-deployment/scripts/add-tags-for-batch-vmss.sh
AhmedAbdoOrtiga: I only mean for adding tags, if we add the resource type as an input, we can use the same script for adding the resourceGroup name as a tag for the vmms

packages/resource-deployment/scripts/recreate-vmss-for-pools.sh
```diff
@@ -0,0 +1,222 @@
...
+    printf " - Running .."
+    while [ $SECONDS -le $endTime ]; do
+
+        local runningCount=$(az batch pool node-counts list \
+                                --query "$nodeTypeContentSelector.running" \
+                                -o tsv
+                            ) 
+        
+        if [[ $runningCount == 0 ]]; then
+            echo "Nodes under $nodeType for pool: $pool under  has started."
```
lamaks: not clear comment. `$nodeType nodes under $pool stopped/deleted?`

packages/resource-deployment/scripts/batch-account-create.sh
lamaks: you can still have it in setup-all- script

packages/resource-deployment/scripts/setup-all-pools-for-batch.sh
lamaks: Still you need fix command name `Enable ...`

packages/resource-deployment/scripts/batch-account-create.sh
devaradhanm: moving it to setup-all-pools-for-batch.sh will require us to pass the secrets to that script, even though it has nothing to do with keyvault. (setup-all-pools doesnt create keyvault.) batch account template is the one which creates the keyvault.

lamaks: I see batch-account-create.sh is only called from install-parallel.sh, the latter had the push-secrets-to-key-vault.sh but beeing moved under former along w/ secrets, so keeping original version possible and no changes required for this pr right?

devaradhanm: I did this change when i was working on enabling monitor. I had to wait for pools to be stable (finish startup task) so that it wont throw lock issues when installing monitoring agent. Waiting for pool nodes will be longer if key vault secrets are not copied, since we have a retry mechanism in startup script. Lets discuss offline, if you still feel this should be outside. I think, if copying secrets should be outside, then so does the batch-pool-setup call.

PR #588, opened 2020-03-11
fix: retry when fetching auth token

### LINKED ISSUE #563: [Bug] Add retries on 500/429 errors while fetching for auth token
The sdk doesnt seem to have retry on failure response(https://github.com/Azure/ms-rest-nodeauth/blob/adea8415de97a8076006fbd9948123d09dfdd78d/lib/credentials/msiAppServiceTokenCredentials.ts#L94). So we may need to add retry & caching, make class singleton for results returned by CredentialsProvider.
 
Refer https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#error-handling

Currently, in Azure function sometimes, we have received the below error in production:
Result: Failure
```
Exception: Error: MSI: Failed to retrieve a token from "http://127.0.0.1:41709/MSI/token/?resource=https%3A%2F%2Fvault.azure.net&api-version=2017-09-01" with an error: {"ExceptionMessage":"","ErrorCode":"service_unavailable","ServiceErrorCodes":null,"StatusCode":503,"Message":null,"CorrelationId":"8b0bb15d-11f7-4a3b-8fbb-3a55ca98a884"}
Stack: Error: MSI: Failed to retrieve a token from "http://127.0.0.1:41709/MSI/token/?resource=https%3A%2F%2Fvault.azure.net&api-version=2017-09-01" with an error: {"ExceptionMessage":"","ErrorCode":"service_unavailable","ServiceErrorCodes":null,"StatusCode":503,"Message":null,"CorrelationId":"8b0bb15d-11f7-4a3b-8fbb-3a55ca98a884"}
    at MSIAppServiceTokenCredentials.<anonymous> (D:\home\site\wwwroot\get-report-func\index.js:56446:23)
    at Generator.next (<anonymous>)
    at fulfilled (D:\home\site\wwwroot\get-report-func\index.js:56384:58)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

## Area

- [ ] Install script
- [ ] Update script
- [ ] Service

pownkel:
#### Description of changes

Add retries when getting auth token in web api client

#### Pull request checklist

### DIFF
--- a/packages/health-client/src/health-checker.ts
+++ b/packages/health-client/src/health-checker.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 import 'reflect-metadata';
 
-import { System } from 'common';
+import { ServiceConfiguration, System } from 'common';
 import { ConsoleLoggerClient, GlobalLogger } from 'logger';
 import { TestRunResult } from 'service-library';
 import { A11yServiceClient, A11yServiceCredential } from 'web-api-client';
@@ -20,28 +20,32 @@ type Argv = {
 };
 
 const argv: Argv = yargs.argv as any;
-const logger = new GlobalLogger([new ConsoleLoggerClient(undefined, console)], process);
-const cred = new A11yServiceCredential(argv.clientId, argv.clientSecret, argv.clientId, argv.authorityUrl, logger);
-const client = new A11yServiceClient(cred, argv.baseUrl);
-const testTimeoutInMinutes = 20;
-const waitTimeBeforeEvaluation = parseInt(argv.waitTimeBeforeEvaluationInMinutes) * 60000;
-const evaluationInterval = parseInt(argv.evaluationIntervalInMinutes) * 60000;
 
 const isTestTimeout = (startTime: Date, currentTime: Date, timeout: number): boolean => {
     return currentTime.getTime() - startTime.getTime() > timeout;
 };
 
 (async () => {
+    const serviceConfig = new ServiceConfiguration();
+    const logger = new GlobalLogger([new ConsoleLoggerClient(serviceConfig, console)], process);
     await logger.setup();
-    console.log('[health-client] Start evaluation of functional tests result.');
-    console.log(`[health-client] Waiting for ${argv.waitTimeBeforeEvaluationInMinutes} minutes before evaluating functional tests result.`);
+    const cred = new A11yServiceCredential(argv.clientId, argv.clientSecret, argv.clientId, argv.authorityUrl, logger);
+    const client = new A11yServiceClient(cred, argv.baseUrl);
+    const testTimeoutInMinutes = 20;
+    const waitTimeBeforeEvaluation = parseInt(argv.waitTimeBeforeEvaluationInMinutes) * 60000;
+    const evaluationInterval = parseInt(argv.evaluationIntervalInMinutes) * 60000;
+
+    logger.logInfo('[health-client] Start evaluation of functional tests result.');
+    logger.logInfo(
+        `[health-client] Waiting for ${argv.waitTimeBeforeEvaluationInMinutes} minutes before evaluating functional tests result.`,
+    );
     await System.wait(waitTimeBeforeEvaluation);
 
     let healthStatus: TestRunResult;
     const startTime = new Date();
     while (healthStatus !== 'pass') {
         try {
-            console.log('[health-client] Retrieving functional tests result.');
+            logger.logInfo('[health-client] Retrieving functional tests result.');
 
             const response = await client.checkHealth(`/release/${argv.releaseId}`);
             if (response.statusCode !== 200) {
@@ -50,21 +54,21 @@ const isTestTimeout = (startTime: Date, currentTime: Date, timeout: number): boo
                 );
             }
 
-            console.log(`[health-client] Functional tests result: ${JSON.stringify(response.body)}`);
+            logger.logInfo(`[health-client] Functional tests result: ${JSON.stringify(response.body)}`);
 
             healthStatus = response.body.healthStatus;
         } catch (error) {
-            console.log(`[health-client] Failed to retrieve functional tests result. ${error}`);
+            logger.logInfo(`[health-client] Failed to retrieve functional tests result. ${error}`);
         }
 
         if (healthStatus !== 'pass') {
             if (isTestTimeout(startTime, new Date(), testTimeoutInMinutes * 60000)) {
-                console.log('[health-client] Functional tests result validation timed out.');
+                logger.logInfo('[health-client] Functional tests result validation timed out.');
 
                 throw new Error('[health-client] Functional tests result validation timed out.');
             }
 
-            console.log(
+            logger.logInfo(
                 `[health-client] Functional tests health status: ${
                     healthStatus ? healthStatus : 'unknown'
                 } . Waiting for next evaluation result.`,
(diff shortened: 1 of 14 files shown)


packages/web-api-client/src/a11y-service-credential.ts
```diff
@@ -40,4 +33,28 @@ export class A11yServiceCredential {
 
         return request.defaults(authOptions);
     }
+
+    private async getTokenWithRetries(numRetries: number): Promise<TokenResponse> {
+        try {
+            return await this.tryGetToken();
+        } catch (err) {
+            if (numRetries > 0) {
+                return this.getTokenWithRetries(numRetries - 1);
```
devaradhanm: can you also add a log here that auth token call failed

packages/web-api-client/src/a11y-service-credential.ts
```diff
@@ -40,4 +33,28 @@ export class A11yServiceCredential {
 
         return request.defaults(authOptions);
     }
+
+    private async getTokenWithRetries(numRetries: number): Promise<TokenResponse> {
+        try {
+            return await this.tryGetToken();
+        } catch (err) {
```
haonliu: Are we doing retries for all kinds of errors, not only 429 and 5xx?

haonliu: And we should wait for at least 1 sec before retry as suggested here: https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#retry-guidance

packages/web-api-client/src/a11y-service-credential.ts
```diff
@@ -40,4 +33,28 @@ export class A11yServiceCredential {
 
         return request.defaults(authOptions);
     }
+
+    private async getTokenWithRetries(numRetries: number): Promise<TokenResponse> {
```
AhmedAbdoOrtiga: I'm a big fan of recursion but I always try to go for iteration for the sake of performance. 
I know in this case the data set or input is small so the difference maybe insignificant, but usually I go for recursion for readability and maintainability, even I believe iteration would be more readable here, so if I were you I'd go for iteration here.
You don't have to change it if you don't want to since as I mentioned it's insignificant in this case but I it just wanted to share my mind.

packages/web-api-client/src/a11y-service-credential.ts
devaradhanm: @pownkel , sorry, i have added a wrong link to fix. We should be fixing it in CredentialProvider/MSICredentialsProvider

packages/web-api-client/src/a11y-service-credential.ts
pownkel: As far as I can tell, there doesn't seem to be an easy way to get the error code from the response. The ErrorResponse object from adal-node only has error and errorDescription (both strings).

haonliu: Error description is created as: 
````
var returnErrorString = operationMessage + ' request returned http error: ' + response.statusCode;
````

Not saying we should parse it, just some thing to consider. I'm fine retry on all errors.

File: packages/health-client/src/health-checker.ts [hunk 4]
```diff
@@ -19,7 +20,8 @@ type Argv = {
 };
 
 const argv: Argv = yargs.argv as any;
-const cred = new A11yServiceCredential(argv.clientId, argv.clientSecret, argv.clientId, argv.authorityUrl);
+const logger = new GlobalLogger([new ConsoleLoggerClient(undefined, console)], process);
+const cred = new A11yServiceCredential(argv.clientId, argv.clientSecret, argv.clientId, argv.authorityUrl, logger);
```
devaradhanm: why are we passing undefined? also why not use dependency injection container like in other places?

packages/web-api-client/src/a11y-service-credential.ts
```diff
@@ -12,22 +13,33 @@ export class A11yServiceCredential {
...
-        return new Promise((resolve, reject) => {
-            this.authContext.acquireTokenWithClientCredentials(this.resourceId, this.clientId, this.clientSecret, (err, tokenResponse) => {
-                if (!isNullOrUndefined(err)) {
-                    reject(err);
-                } else {
-                    resolve(tokenResponse as TokenResponse);
+        await this.logger.setup();
+
+        let lastError: Error;
+        for (let i = 0; i < this.maxTokenAttempts; i += 1) {
```
devaradhanm: create a utility file, may be RetryHelper, that is generic & can be used in other places

packages/health-client/src/health-checker.ts
pownkel: This is the script we use for E2E validation in the release pipeline, which doesn't have a dependency injection container. The undefined value is for the ServiceConfiguration. Since we're only using one configuration value to enable/disable logging, and we don't use the service config anywhere else in the script, it seemed like overkill to create one just for this.

packages/web-api-client/src/a11y-service-credential.ts
```diff
@@ -12,22 +13,33 @@ export class A11yServiceCredential {
...
     }
 
     public async getToken(): Promise<TokenResponse> {
-        return new Promise((resolve, reject) => {
-            this.authContext.acquireTokenWithClientCredentials(this.resourceId, this.clientId, this.clientSecret, (err, tokenResponse) => {
-                if (!isNullOrUndefined(err)) {
-                    reject(err);
-                } else {
-                    resolve(tokenResponse as TokenResponse);
+        await this.logger.setup();
```
devaradhanm: logger setup should not be in this class. This class should expect that the logger is already setup.

packages/health-client/src/health-checker.ts
devaradhanm: why not move all this inside the async function.
& create other instances after logger.setup is called.
& use logger for logging instead of console.log in this file

packages/logger/src/base-console-logger-client.ts
```diff
@@ -27,7 +27,12 @@ export abstract class BaseConsoleLoggerClient implements LoggerClient {
     public async setup(baseProperties?: BaseTelemetryProperties): Promise<void> {
         this.baseProperties = baseProperties;
 
-        this.isConsoleLogEnabled = (await this.serviceConfig.getConfigValue('logConfig')).logInConsole;
+        if (this.serviceConfig !== undefined) {
```
devaradhanm: use lodash isNil

devaradhanm: also, why will this ever be null? if service config is not found, would we be using the default runtime-config?

packages/web-api-client/src/a11y-service-credential.ts
```diff
@@ -30,14 +59,16 @@ export class A11yServiceCredential {
...
-            headers: {
-                authorization: `${accessToken.tokenType} ${accessToken.accessToken}`,
-            },
-        };
+    private async handleGetTokenError(err: Error): Promise<void> {
+        this.logger.logError(`Auth getToken call failed with error: ${JSON.stringify(err)}`);
 
-        return request.defaults(authOptions);
+        if (this.sleepBetweenRetries) {
+            await this.sleep(1000);
```
devaradhanm: this code should be in Retry helper.
RetryHelper.retry<T>(callback: ()=> T, maxAttempts, retryIntervalInMilliSec): Promise<T>

packages/logger/src/base-console-logger-client.ts
pownkel: It would be null in the health-checker script, because we don't have a ServiceConfiguration in that file, and I didn't create one because we would only be using one value anyway to enable/disable console logging. Should we be creating a service config in health-checker?

PR #591, opened 2020-03-12
fix: wait for locks before configuring dpkg

devaradhanm:
#### Description of changes

#### Pull request checklist

### DIFF
--- a/packages/resource-deployment/scripts/pool-startup/pool-startup-internal.sh
+++ b/packages/resource-deployment/scripts/pool-startup/pool-startup-internal.sh
@@ -7,15 +7,15 @@
 
 set -eo pipefail
 
-waitForAptUpdates() {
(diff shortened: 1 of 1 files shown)


packages/resource-deployment/scripts/pool-startup/pool-startup-internal.sh
```diff
@@ -15,6 +15,8 @@ waitForAptUpdates() {
     done
 }
 
+waitForAptUpdates
```
devaradhanm: the very first time when the node starts, dpkg configuration fails sometimes due to lock issues.

AhmedAbdoOrtiga: since the function is used to wait for locks other than "apt-get update", should we change the function name?

PR #592, opened 2020-03-12
chore: Refactor create function app

AhmedAbdoOrtiga:
#### Description of changes
- Merge deploy arm template code for web worker and wep api in one function.

#### Pull request checklist

### DIFF
--- a/packages/resource-deployment/scripts/function-app-create.sh
+++ b/packages/resource-deployment/scripts/function-app-create.sh
@@ -85,7 +85,7 @@ publishFunctionAppScripts() {
     # Run function tool with retries due to app service warm up time delay
     end=$((SECONDS + 120))
     while [ $SECONDS -le $end ]; do
-        func azure functionapp publish $functionAppName --node || true
+         func azure functionapp publish $functionAppName --node || true
         if [ $? -eq 0 ]; then
             break
         fi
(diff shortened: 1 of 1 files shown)


packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -123,58 +123,30 @@ getFunctionAppPrincipalId() {
...
+    myWebFunctionAppName="$resourceName"
 
-    waitForFunctionAppServiceDeploymentCompletion $webApiFunctionAppName
-    echo "Successfully deployed Azure Function App '$webApiFunctionAppName'"
-}
-
-deployWebWorkersArmTemplate() {
-    functionAppNamePrefix="web-workers-allyfuncapp"
-    templateFilePath="${0%/*}/../templates/function-web-workers-app-template.json"
+    waitForFunctionAppServiceDeploymentCompletion $myWebFunctionAppName
```
devaradhanm: just name it functionAppName

packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -109,72 +109,44 @@ publishFunctionAppScripts() {
...
     . "${0%/*}/key-vault-enable-msi.sh"
-}
 
-deployWebWorkersFunctionApp() {
-    deployWebWorkersArmTemplate $webWorkersPackageName
-
-    # Keep child script call only one level deep to preserve exports
-    getFunctionAppPrincipalId $webWorkersFunctionAppName
-    . "${0%/*}/key-vault-enable-msi.sh"
+    eval $functionAppName="'$myFunctionAppName'"
```
devaradhanm: why not just functionAppName="'$myFunctionAppName'" ?

devaradhanm: also why is this line needed? who is using it?

AhmedAbdoOrtiga: I don't want to set functionAppName, I want to set it's value, this is one way to handle Returning Values from Bash Functions
so the command (eval $functionAppName="'$myFunctionAppName'") will be compiled as 
(eval webAPiFunctionAppName="'$myFunctionAppName'") for webApi call and as
(eval webWorkerFunctionAppName="'$myFunctionAppName'") for webWorker call

Look at this: https://www.linuxjournal.com/content/return-values-bash-functions

devaradhanm: P2: i see. you could use get-resource-names.sh to get the names.

PR #597, opened 2020-03-13
fix: fix bug in total items added to queue.

devaradhanm:
#### Description of changes
we were always adding items to queue in multiples of 100. So, even if we set maxQueueSize to 20, we will end up adding 100

#### Pull request checklist

### DIFF
--- a/packages/web-api-scan-request-sender/src/sender/on-demand-dispatcher.ts
+++ b/packages/web-api-scan-request-sender/src/sender/on-demand-dispatcher.ts
@@ -32,9 +32,11 @@ export class OnDemandDispatcher {
         let continuationToken;
 
         do {
+            const totalItemsToAdd = configQueueSize - currentQueueSize;
             do {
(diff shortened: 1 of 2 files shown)


packages/web-api-scan-request-sender/src/sender/on-demand-dispatcher.ts
```diff
@@ -32,9 +32,11 @@ export class OnDemandDispatcher {
         let continuationToken;
 
         do {
```
AhmedAbdoOrtiga: Do we need this outer loop after this change? 
we are going to scan totalItemsToAdd (or less if the query returned less) all at once in the inner loop or it will return no items

devaradhanm: as discussed offline, we dont need both the loops. but that needs to be fixed once we refactor the cosmoscontainerclient.readItems. not taking that as part of this pr.

PR #599, opened 2020-03-13
fix: create keyvault only if it doesn't exist.

devaradhanm:
- moved keyvault creation out of batch setup to a new file 
- use ARM to create keyvault only if it doesn't exist.
- made more processes to run in parallel; (check install-parallel file)
- added script to kill all child processes if there is any error. (check process-utils & install-parallel file)
- added script to wait for pending deployments before starting deployment. (wait-for-pending-deployments.sh)
- removed redundant args being passed via command line in more files to make it easier for testing.

#### Description of changes

#### Pull request checklist


packages/resource-deployment/scripts/process-utilities.sh
```diff
@@ -12,20 +12,69 @@ function waitForProcesses() {
     for pid in "${!list}"; do
         echo "Waiting for process with pid $pid"
         wait $pid
-        echo "Process with pid $pid exited"
+        local pExitCode=$?
+        echo "Process with pid $pid exited with exit code $pExitCode"
+
+        if [[ $pExitCode != 0 ]]; then
+            echo "Process - $pid failed with exit code $pExitCode. Killing current proccess."
+            exit $pExitCode
```
lamaks: where do we kill process in this func?

packages/resource-deployment/scripts/process-utilities.sh
```diff
@@ -12,20 +12,69 @@ function waitForProcesses() {
...
+
+function sendSignalToProcessIfExists {
+    local currentPid=$1	
+    local signal=$2
+
+    if [[ -z $currentPid ]]; then	
+        return	
+    fi	
+
+    if kill -0 $currentPid > /dev/null 2>&1; then
```
lamaks: when we kill arm template related process azure will keep deployment running and we may need to go to azure to delete a deployment; otherwise it can take much longer than normal deployment flow, example - api management deployment cannot be re-run in this scenario

build.yaml
```diff
@@ -17,22 +17,25 @@ steps:
...
+    # - script: |
+    #       yarn copyright:check
+    #   displayName: check for copyright headers
 
-    - script: |
-          yarn clean
-      displayName: clean
+    # - script: |
+    #       yarn clean
+    #   displayName: clean
```
lamaks: temp?

packages/resource-deployment/templates/batch-account.template.json
```diff
@@ -30,88 +30,23 @@
                 "description": "Storage account name."
             }
         },
-        "logsRetentionInDays": {
```
lamaks: do we have replacement for this in a script?

packages/resource-deployment/scripts/install-parallel.sh
```diff
@@ -74,6 +75,28 @@ Azure region - Azure region where the instances will be deployed. Available Azur
 
 . "${0%/*}/process-utilities.sh"
 
+function onError() {
+    echo "insiden on error"
```
lamaks: typo

packages/resource-deployment/scripts/process-utilities.sh
```diff
@@ -12,20 +12,69 @@ function waitForProcesses() {
...
+{	
+    local currentPid=$1	
+
+    if [[ -z $currentPid ]]; then	
+        return	
+    fi	
+
+    if kill -0 $currentPid > /dev/null 2>&1; then	
+        echo "stopping process $currentPid"	
+        sendSignalToProcessIfExists $currentPid	-STOP
```
lamaks: is -stop a valid signal?

packages/resource-deployment/scripts/process-utilities.sh
```diff
@@ -12,20 +12,69 @@ function waitForProcesses() {
...
+       
+        killDescendentProcesses  $currentPid
+       
+        sendSignalToProcessIfExists $currentPid	-9
```
lamaks: seems tab used
add comment what is signal used

packages/resource-deployment/scripts/process-utilities.sh
devaradhanm: yep. aware of this issue. May be i can print that in the console log.

build.yaml
devaradhanm: yes.

packages/resource-deployment/scripts/process-utilities.sh
devaradhanm: yes. to suspend process, so that it wont produce new children when i am fetching its children count 😄

packages/resource-deployment/scripts/process-utilities.sh
devaradhanm: not here. it is used in install_paralle.

packages/resource-deployment/scripts/process-utilities.sh
devaradhanm: prefixing SIG is optional for all signals

packages/resource-deployment/scripts/process-utilities.sh
devaradhanm: not sure why it shows like this. but there is no tab. just a single space

packages/resource-deployment/scripts/process-utilities.sh
devaradhanm: it is not supported in az cli https://github.com/Azure/azure-cli/issues/12252

lamaks: seems not implemented yet

devaradhanm: added WARN log in onExit function in install_parallel.

packages/resource-deployment/scripts/process-utilities.sh
lamaks: still message is confusing here `Killing current proccess.` also typo here

devaradhanm: i can change the message if u prefer. but, i don;t see typo.

packages/resource-deployment/scripts/process-utilities.sh
lamaks: We need to postpone the change with kill implementation till we have reliable way to terminate pending deployment on Azure

packages/resource-deployment/scripts/process-utilities.sh
lamaks: yes

Issue #601, opened 2020-03-14
[Bug] Fix unit tests that take > 100 ms


devaradhanm:
## Describe the bug
Unit tests should not take more than 100 ms to execute.
Currently we have around 600 tests & it takes 6-8 min to run them.
We have some tests that take > 500 ms time due to waits. waits have to be mocked out.

You can use https://www.npmjs.com/package/jest-slow-test-reporter & https://github.com/microsoft/accessibility-insights-service/pull/602 to find the slow running tests

some of the tests that take > 100ms:
```
Top 20 slowest examples (3.998 seconds, 51.8% of total time):
  CosmosContainerClient.tryExecuteOperation() invoke operation callback with timeout
    1.006 seconds ./packages/azure-services/src/storage/cosmos-container-client.spec.ts
  processWebRequest handles request
    0.928 seconds ./packages/web-workers/src/process-web-request.spec.ts
  processWebRequest Returns response from controller
    0.873 seconds ./packages/web-api/src/process-request.spec.ts
  CosmosContainerClient.tryExecuteOperation() invoke operation callback with retry
    0.403 seconds ./packages/azure-services/src/storage/cosmos-container-client.spec.ts
  Batch createTasks() should add no more than 100 tasks in a single Batch API call
    0.088 seconds ./packages/azure-services/src/azure-batch/batch.spec.ts
  SecretProvider gets secret
    0.065 seconds ./packages/azure-services/src/key-vault/secret-provider.spec.ts
  WhyNodeRunningLogger logAfterSeconds invoke log after timeout
    0.064 seconds ./packages/common/src/why-node-running-logger.spec.ts
  AIScanner should close browser if exception occurs
    0.063 seconds ./packages/cli/src/scanner/ai-scanner.spec.ts
  Command Runner scan with output parameter specified output directory exists
    0.057 seconds ./packages/cli/src/command-runner.spec.ts
  OrchestrationStepsImpl callHealthCheckActivity healthCheckActivity throws error on status code 199
    0.054 seconds ./packages/web-workers/src/orchestration-steps.spec.ts
  ScanReportController handleRequest should return 400 if request id is invalid
    0.046 seconds ./packages/web-api/src/controllers/scan-report-controller.spec.ts
  AIScanner should launch browser page with given url and scan the page with axe-core
    0.045 seconds ./packages/cli/src/scanner/ai-scanner.spec.ts
  BatchScanResultController handleRequest should return different response for different kind of scanIds
[...]

PR #602, opened 2020-03-14
chore: Publish test results to pipeline & generate html test report

devaradhanm:
#### Description of changes

- Publishing test results so that we can see the tests in azure devops.
- Generating html report to view the tests & time taken locally.
- Added commented code to get the slow running test information.

#### Pull request checklist


build.yaml
```diff
@@ -42,6 +42,13 @@ steps:
           yarn test --ci --coverage
       displayName: run unit tests
 
+    - task: PublishTestResults@2
+      inputs:
+          testResultsFiles: $(System.DefaultWorkingDirectory)/test-results/unit/junit.xml
+          testRunTitle: $(Agent.JobName)
+      condition: always()
```
AhmedAbdoOrtiga: does this mean it will run even if the "build" task or the "run test" task failed, if so we might need to remove this condition

jest.config.base.js
```diff
@@ -37,7 +37,31 @@ module.exports = {
...
+        ],
+        [
+            'jest-html-reporter',
+            {
+                pageTitle: 'html report',
+                outputPath: './test-results/unit/html-reporter/report.html',
+            },
+        ],
+        // [
+        //     'jest-slow-test-reporter',
```
AhmedAbdoOrtiga: Why do we need to keep this?

package.json
```diff
@@ -33,7 +33,9 @@
         "jest": "^25.1.0",
         "jest-circus": "^25.1.0",
         "jest-extended": "^0.11.2",
+        "jest-html-reporter": "^2.8.0",
         "jest-junit": "^10.0.0",
+        "jest-slow-test-reporter": "^1.0.0",
```
AhmedAbdoOrtiga: do we need this?

build.yaml
devaradhanm: yes. it should. if the test fails, we still want to publish the results.

package.json
devaradhanm: it is used by the commented code that i have.

jest.config.base.js
devaradhanm: we shouldnt enable this as it increases test run time. But, if we want to find slow running tests, we just need to uncomment this & run tests. to fix this issue (https://github.com/microsoft/accessibility-insights-service/issues/601)

build.yaml
AhmedAbdoOrtiga: I mean the task fails, the build task, there'll be nothing to publish, right?

package.json
AhmedAbdoOrtiga: my point that the code is commented, so technically we don't use it, but it's ok

build.yaml
devaradhanm: yes. in that case, it won't publish  anything. check web repo. we have the same code there as well. If we are not publishing test results as part of running tests, we have to run the step always.

package.json
devaradhanm: we have lots of long running tests now. We can remove it once we fix the issue i created. I can may be added a comment.

PR #609, opened 2020-03-17
feat: add notification state to the scan document interface

haonliu:
Adding notification state to the scan document interface

### DIFF
--- a/packages/storage-documents/src/on-demand-page-scan-result.ts
+++ b/packages/storage-documents/src/on-demand-page-scan-result.ts
@@ -37,7 +37,18 @@ export interface OnDemandPageScanResult extends StorageDocument {
     priority: number;
     itemType: ItemType.onDemandPageScanRunResult;
     batchRequestId: string;
-    notificationState?: NotificationState;
+    notification?: ScanCompletedNotification;
+}
+
+export interface ScanCompletedNotification {
+    state: NotificationState;
+    notificationUrl: string;
+    error: NotificationError;
+}
+
+export interface NotificationError {
+    errorCode: number;
+    message: string;
 }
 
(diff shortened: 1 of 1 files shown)


packages/storage-documents/src/on-demand-page-scan-result.ts
```diff
@@ -36,6 +37,7 @@ export interface OnDemandPageScanResult extends StorageDocument {
     priority: number;
     itemType: ItemType.onDemandPageScanRunResult;
     batchRequestId: string;
+    notificationState?: NotificationState;
```
devaradhanm: we should have an object that contains things like notification status, error message, error code etc

lamaks: > notification status, error message, error code etc
are not candidates for the single object, why we need that object? Example?

devaradhanm: pageScanResult {
run: {...},
notification: {url: string, status: enum, error: {code, message}}
}

packages/storage-documents/src/on-demand-page-scan-result.ts
```diff
@@ -36,6 +37,18 @@ export interface OnDemandPageScanResult extends StorageDocument {
...
+}
+
+export interface ScanCompletedNotification {
+    notificationUrl: string;
+    state?: NotificationState;
+    error?: NotificationError;
+}
+
+export interface NotificationError {
+    errorType: string;
```
devaradhanm: these should be in one of the known strings. (like ScanError). If not sure what all notification error types we will need, then just add one
NotificationErrorTypes= "HttpErrorCode"

PR #614, opened 2020-03-17
feat (send-notification): update post scan api to accept reply url

haonliu:
Accept reply url sending with the request and save it into the database. 

Will check if the reply url is valid if the user does provide it. If not, will return an error response.

Api version management refactor will be a separate PR. We might need to put that in the service config.

### DIFF
--- a/packages/service-library/src/web-api/api-contracts/scan-run-request.ts
+++ b/packages/service-library/src/web-api/api-contracts/scan-run-request.ts
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 export interface ScanRunRequest {
     url: string;
-
+    replyUrl?: string;
     /**
      * Priority values can range from -1000 to 1000, with -1000 being the lowest priority and 1000 being the highest priority.
      * The default value is 0.
--- a/packages/web-api/src/controllers/scan-request-controller.spec.ts
+++ b/packages/web-api/src/controllers/scan-request-controller.spec.ts
@@ -74,14 +74,14 @@ describe(ScanRequestController, () => {
     }
 
     describe(ScanRequestController, () => {
-        // it('rejects request with large payload', async () => {
-        //     context.req.rawBody = JSON.stringify([{ url: '' }, { url: '' }, { url: '' }, { url: '' }]);
-        //     scanRequestController = createScanRequestController(context);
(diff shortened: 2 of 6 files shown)


packages/web-api/src/controllers/scan-request-controller.ts
```diff
@@ -88,6 +85,7 @@ export class ScanRequestController extends ApiController {
         this.logger.trackEvent('BatchScanRequestSubmitted', null, measurements);
     }
 
+    // tslint:disable-next-line: no-any
```
devaradhanm: avoid no-any in product code. add strict typings

packages/web-api/src/controllers/scan-request-controller.spec.ts
```diff
@@ -74,13 +74,22 @@ describe(ScanRequestController, () => {
     }
 
     describe(ScanRequestController, () => {
-        it('rejects request with large payload', async () => {
-            context.req.rawBody = JSON.stringify([{ url: '' }, { url: '' }, { url: '' }, { url: '' }]);
+        // it('rejects request with large payload', async () => {
```
devaradhanm: remove commented code

packages/service-library/src/web-api/api-contracts/scan-run-request.ts
```diff
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 export interface ScanRunRequest {
     url: string;
-
+    replyUrl?: string;
```
manishsat: I think WCP team sends {url, priority} and they will send  {url, priority, replyUrl}, plz verify that the sequence and I am happy to know whether sequence matters or not.

packages/web-api/src/controllers/scan-request-controller.spec.ts
manishsat: +1

devaradhanm:
update pr title to feat (send-notification): ....

packages/service-library/src/web-api/api-contracts/scan-run-request.ts
haonliu: I don't think the order matters, cause we are parsing them as JSON object.

PR #616, opened 2020-03-18
fix: return InvalidResourceId on invalid guid

### LINKED ISSUE #458: [Bug] error response code not as expected.
## Describe the bug

The rest api always returns 404 with no err message when any error occurs.

## Area

- [ ] Install script
- [ ] Update script
- [x] Service

## To Reproduce

send a scan status get request with an invalid guid.

## Expected behavior

the response should indicate that the input is not valid.

pownkel:
#### Description of changes

Previously we were throwing 404 if the user asked for scan status or report with anything not formatted as a guid. This will allow us to return an error message instead, as was intended.

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #458
- [x] Added relevant ~~unit test~~ functional test for your changes. (`yarn test`)

### DIFF
--- a/packages/web-api/get-report-func/function.json
+++ b/packages/web-api/get-report-func/function.json
@@ -5,7 +5,7 @@
             "name": "request",
             "type": "httpTrigger",
             "direction": "in",
-            "route": "scans/{scanId:guid}/reports/{reportId:guid}",
+            "route": "scans/{scanId}/reports/{reportId}",
             "methods": ["get"]
         },
         {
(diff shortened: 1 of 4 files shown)


packages/web-api/get-report-func/function.json
```diff
@@ -5,7 +5,7 @@
             "name": "request",
             "type": "httpTrigger",
             "direction": "in",
-            "route": "scans/{scanId:guid}/reports/{reportId:guid}",
+            "route": "scans/{scanId}/reports/{reportId}",
```
devaradhanm: what was the experience before & after this change?
how does azure function handle if scan id was not guid before?
if we make this change, we need to do scan id & report id type check in code. are we doing this?

pownkel: the code already checks if the guids are valid, but if we add the guid type in the http route then the function app is only triggered if the id is formatted like a guid. So right now, we have the guid check in code, but that code only runs if we get a valid guid in the first place. After the change, the function app will get triggered whatever the scanId is, and then we can check the scanId and send an error message if the check fails.

PR #619, opened 2020-03-19
feat (send notification): add task config generator for send notification

devaradhanm:
#### Description of changes

Refactored https://github.com/microsoft/accessibility-insights-service/blob/master/packages/web-api-scan-job-manager/src/batch/scanner-batch-task-config.ts &  https://github.com/microsoft/accessibility-insights-service/blob/master/packages/web-api-scan-job-manager/src/batch/scan-task-parameter-provider.ts and created the below two files moving them to azure-services package.
- Added BatchTaskConfigGenerator(batch-task-config-generator.ts) which generates the task configuration
- Added abstract BatchTaskPropertyProvider that provides task specific property values for task config generation. (batch-task-property-provider.ts) web-api-scan-job-manager & notification job manager extends this class to provide task property values. 

Created EnvironmentSettings class to access environment instead of directly using process.env. Currently just using this for config generator only.

#### Pull request checklist


packages/web-api-send-notification-job-manager/src/task/send-notification-task-property-provider.ts
```diff
@@ -0,0 +1,33 @@
...
+import { EnvironmentSettings, ServiceConfiguration } from 'common';
+import { inject, injectable } from 'inversify';
+
+@injectable()
+export class SendNotificationTaskPropertyProvider extends BatchTaskPropertyProvider {
+    constructor(
+        @inject(EnvironmentSettings) private readonly environmentSettings: EnvironmentSettings,
+        @inject(ServiceConfiguration) serviceConfig: ServiceConfiguration,
+    ) {
+        super(serviceConfig, 'start-web-api-send-notification-runner.sh', ['id', 'url']);
```
devaradhanm: we need to change this properties based on what information we have in the queue. will update this list in the next PR

packages/common/src/system/environment-settings.spec.ts
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import 'reflect-metadata';
+
+import { inject } from 'inversify';
```
AhmedAbdoOrtiga: remove unused imports

packages/service-library/package.json
```diff
@@ -39,6 +37,8 @@
         "typescript": "^3.8.3"
     },
     "dependencies": {
+        "@azure/cosmos": "^3.6.2",
```
AhmedAbdoOrtiga: is there a reason to move it from dev-dependencies to dependencies?

devaradhanm: because these are not dev dependencies. (dev dependencies are the dependencies you use at compile time. like typings, webpack etc.

PR #635, opened 2020-03-23
fix: update scan run without reading first

### LINKED ISSUE #620: [Bug] Remove unused data from the queue message
## Describe the bug
- In the scan request sender, we add id, url and priority to the queued message.
- But actually we only need the id, we have to read the document and for that the id is enough.
- The url and the priority aren't used anywhere.
- It'd be better if we remove them.

Also try to use merge for the document instead of reading then updating

Maybe something like 

public createPartitionKeyForDocument(documentType: ItemType, documentId: string) to get PK so id & doc type should be enough 

publicasync mergeOrWriteDocument<T extends CosmosDocument>(document: T, partitionKey?: string) then can handle client side merge for now

## Area

- [ ] Install script
- [ ] Update script
- [ ] Service

pownkel:
#### Description of changes

- Remove unused priority field from OnDemandScanRequestMessage
- Add batchRequestId to OnDemandScanRequestMessage
- Refactor updateScanRuns to accept a partial document (with only id required)
- Refactor runner to update the scan run document without first reading it

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #620 

### DIFF
--- a/packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
+++ b/packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
@@ -13,7 +13,7 @@ export class OnDemandPageScanRunResultProvider {
         @inject(cosmosContainerClientTypes.OnDemandScanRunsCosmosContainerClient)
         private readonly cosmosContainerClient: CosmosContainerClient,
         @inject(PartitionKeyFactory) private readonly partitionKeyFactory: PartitionKeyFactory,
-    ) { }
+    ) {}
 
     public async readScanRun(scanId: string): Promise<OnDemandPageScanResult> {
         return (await this.cosmosContainerClient.readDocument<OnDemandPageScanResult>(scanId, this.getPartitionKey(scanId))).item;
--- a/packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
+++ b/packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
@@ -4,7 +4,7 @@ import { inject, injectable } from 'inversify';
 
 import { CosmosContainerClient, cosmosContainerClientTypes } from 'azure-services';
 import { flatMap, groupBy } from 'lodash';
-import { ItemType, OnDemandPageScanResult } from 'storage-documents';
+import { ItemType, OnDemandPageScanResult, PartialOnDemandPageScanResult } from 'storage-documents';
 import { PartitionKeyFactory } from '../factories/partition-key-factory';
 
 @injectable()
@@ -45,10 +45,11 @@ export class OnDemandPageScanRunResultProvider {
         return flatMap(response);
     }
 
-    public async updateScanRun(pageScanResult: OnDemandPageScanResult): Promise<OnDemandPageScanResult> {
-        this.setSystemProperties(pageScanResult);
+    public async updateScanRun(pageScanResult: PartialOnDemandPageScanResult): Promise<OnDemandPageScanResult> {
+        const storableResult = pageScanResult as OnDemandPageScanResult;
+        this.setSystemProperties(storableResult);
 
-        return (await this.cosmosContainerClient.mergeOrWriteDocument(pageScanResult)).item;
+        return (await this.cosmosContainerClient.mergeOrWriteDocument(storableResult)).item;
     }
 
     public async writeScanRuns(scanRuns: OnDemandPageScanResult[]): Promise<void> {
--- a/packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
+++ b/packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
@@ -46,10 +46,10 @@ export class OnDemandPageScanRunResultProvider {
     }
 
     public async updateScanRun(pageScanResult: PartialOnDemandPageScanResult): Promise<OnDemandPageScanResult> {
-        const storableResult = pageScanResult as OnDemandPageScanResult;
-        this.setSystemProperties(storableResult);
+        const persistedResult = pageScanResult as OnDemandPageScanResult;
+        this.setSystemProperties(persistedResult);
 
-        return (await this.cosmosContainerClient.mergeOrWriteDocument(storableResult)).item;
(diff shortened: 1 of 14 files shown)


packages/storage-documents/src/on-demand-scan-request-message.ts
```diff
@@ -3,6 +3,6 @@
 
 export interface OnDemandScanRequestMessage {
     id: string;
+    batchRequestId: string;
```
devaradhanm: why are we adding batchRequestId?

packages/storage-documents/src/on-demand-page-scan-result.ts
```diff
@@ -71,3 +71,7 @@ export interface OnDemandPageScanRunResult {
     pageTitle?: string;
     pageResponseCode?: number;
 }
+
+export interface PartialOnDemandPageScanResult extends Partial<OnDemandPageScanResult> {
+    id: string;
```
devaradhanm: also add partitionKey

devaradhanm: these two should be mandatory fields to avoid cross partition query

packages/storage-documents/src/on-demand-scan-request-message.ts
pownkel: Originally we got batchRequestId from reading the scan run, which we aren't doing anymore

packages/storage-documents/src/on-demand-page-scan-result.ts
pownkel: We can get the partitionKey using the id and itemType, so I don't think we need to require it here

packages/storage-documents/src/on-demand-page-scan-result.ts
```diff
@@ -71,3 +71,7 @@ export interface OnDemandPageScanRunResult {
     pageTitle?: string;
     pageResponseCode?: number;
 }
+
+export interface PartialOnDemandPageScanResult extends Partial<OnDemandPageScanResult> {
```
lamaks: likely this interface does not require any members at all since it should provide all `OnDemandPageScanResult` members per `Partial` target functionality

packages/service-library/src/data-providers/on-demand-page-scan-run-result-provider.ts
```diff
@@ -45,10 +45,11 @@ export class OnDemandPageScanRunResultProvider {
         return flatMap(response);
     }
 
-    public async updateScanRun(pageScanResult: OnDemandPageScanResult): Promise<OnDemandPageScanResult> {
-        this.setSystemProperties(pageScanResult);
+    public async updateScanRun(pageScanResult: PartialOnDemandPageScanResult): Promise<OnDemandPageScanResult> {
+        const storableResult = pageScanResult as OnDemandPageScanResult;
```
lamaks: may be `persistedResult`

packages/storage-documents/src/on-demand-page-scan-result.ts
pownkel: I think we do need to require id, since we need it to get the partitionKey and update the document

packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -37,25 +37,31 @@ export class Runner {
...
-        pageScanResult = await this.onDemandPageScanRunResultProvider.readScanRun(scanMetadata.id);
-        this.logger.setCustomProperties({ scanId: scanMetadata.id, batchRequestId: pageScanResult.batchRequestId });
+        this.logger.setCustomProperties({ scanId: scanMetadata.id, batchRequestId: scanMetadata.batchRequestId });
 
         this.logger.trackEvent('ScanTaskStarted', undefined, { scanWaitTime: (scanStartedTimestamp - scanSubmittedTimestamp) / 1000 });
 
         this.logger.logInfo(`Updating page scan run result state to running.`);
-        pageScanResult = this.resetPageScanResultState(pageScanResult);
-        pageScanResult = await this.onDemandPageScanRunResultProvider.updateScanRun(pageScanResult);
+        const pageScanResult: PartialOnDemandPageScanResult = {
```
lamaks: probably we do not need that strong `PartialOnDemandPageScanResult` type, having `Partial<OnDemandPageScanResult>` on `onDemandPageScanRunResultProvider.updateScanRun()` definition will suffice

packages/storage-documents/src/on-demand-page-scan-result.ts
lamaks: As per other comment I'll drop this type. `Partial<>` as func param type should let use `OnDemandPageScanResult` arbitrary members

packages/storage-documents/src/on-demand-scan-request-message.ts
devaradhanm: we have deprecated batch calls. so we dont need this anymore.

packages/storage-documents/src/on-demand-page-scan-result.ts
devaradhanm: you need to validate just querying for id doesnt result in cross partition query

packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -37,25 +37,31 @@ export class Runner {
...
+        this.logger.setCustomProperties({ scanId: scanMetadata.id });
 
         this.logger.trackEvent('ScanTaskStarted', undefined, { scanWaitTime: (scanStartedTimestamp - scanSubmittedTimestamp) / 1000 });
 
         this.logger.logInfo(`Updating page scan run result state to running.`);
-        pageScanResult = this.resetPageScanResultState(pageScanResult);
-        pageScanResult = await this.onDemandPageScanRunResultProvider.updateScanRun(pageScanResult);
+        const pageScanResult: Partial<OnDemandPageScanResult> = {
+            id: scanMetadata.id,
+            run: {
```
lamaks: we may need to keep the following to reset data in a storage:
`      originPageScanResult.scanResult = null;	
        originPageScanResult.reports = null;`

PR #637, opened 2020-03-24
fix: reduce cosmos container throughput

### LINKED ISSUE #626: [Bug] Reduce cosmos throughput
## Describe the bug
We can reduce some of our containers throughput both in prod & in dev environment.

For eg:
systemData container is only used by job manager task (1 instance). Currently this is set as 2k throughput unit. This can be brought down to 400 in all environments

a11yIssues - This is used for the legacy flow. & is not being used in dev environment. this can be brought down to 400 for dev. For production, we need to check the traffic.

All other containers should be brought down to 400, since we don't have much traffic  & our retry logic should be enough to handle cosmos throughput issues for dev environment.

This change will bring down the cost from **19.2$ per day(134.4$ per week)**  per resource group to **3.85$ per day (26.95 per week)** 

## Area

- [ ] Install script
- [ ] Update script
- [ ] Service

pownkel:
#### Description of changes

Reduce cosmos container throughput to reduce cost for dev resource groups

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #626

### DIFF
--- a/packages/resource-deployment/scripts/setup-cosmos-db.sh
+++ b/packages/resource-deployment/scripts/setup-cosmos-db.sh
@@ -72,21 +72,21 @@ function setupCosmos() {
 
     # Increase throughput for below collection only in case of prod
     # Refer to https://docs.microsoft.com/en-us/azure/cosmos-db/time-to-live for item TTL scenarios
-    if [ $environment = "prod" ]; then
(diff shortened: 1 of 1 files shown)


devaradhanm:
LGTM. but, have you validated this change?

PR #639, opened 2020-03-24
feat(send-notification): adding scan notification task creator

devaradhanm:
#### Description of changes
refactored web-api-job-manager's Worker to use the base class BatchTaskCreator.
This pr also fixes not to add more tasks than the maximum allowed running tasks. 

#### Pull request checklist

### DIFF
--- a/packages/azure-services/src/azure-batch/pool-load-generator.spec.ts
+++ b/packages/azure-services/src/azure-batch/pool-load-generator.spec.ts
@@ -2,12 +2,12 @@
 // Licensed under the MIT License.
 import 'reflect-metadata';
 
-import { ServiceConfiguration } from 'common';
+import { JobManagerConfig, ServiceConfiguration } from 'common';
 import * as moment from 'moment';
 import { IMock, Mock } from 'typemoq';
 import { PoolLoadGenerator, PoolMetricsInfo } from './pool-load-generator';
 
-// tslint:disable: no-unsafe-any
+// tslint:disable: no-unsafe-any no-object-literal-type-assertion
 
 let poolMetricsInfo: PoolMetricsInfo;
 let poolLoadGenerator: PoolLoadGenerator;
@@ -26,7 +26,7 @@ describe(PoolLoadGenerator, () => {
                     activeToRunningTasksRatio: activeToRunningTasksRatio,
                     addTasksIntervalInSeconds: 15,
                     maxWallClockTimeInHours: 1,
-                };
+                } as JobManagerConfig;
             });
 
         jest.spyOn(process, 'hrtime').mockImplementation((time?: [number, number]) => [5, 0]);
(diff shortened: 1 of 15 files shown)


packages/common/src/configuration/service-configuration.ts
```diff
@@ -4,6 +4,7 @@ import * as convict from 'convict';
 import * as fs from 'fs';
 import { injectable } from 'inversify';
 import { isNil } from 'lodash';
+import { format } from 'url';
```
AhmedAbdoOrtiga: remove unused import

packages/job-manager/src/batch/pool-load-generator.spec.ts
```diff
@@ -22,7 +24,7 @@ describe(PoolLoadGenerator, () => {
                     activeToRunningTasksRatio: activeToRunningTasksRatio,
                     addTasksIntervalInSeconds: 15,
                     maxWallClockTimeInHours: 1,
-                };
+                } as JobManagerConfig;
```
AhmedAbdoOrtiga: can you add a value for "sendNotificationTasksCount" instead of casting?, and then check the value in the UTs something like: **expect(poolLoadGenerator.sendNotificationTasksCount).toEqual(x);**

packages/job-manager/src/worker/worker.spec.ts
```diff
@@ -63,7 +63,7 @@ describe(Worker, () => {
                     activeToRunningTasksRatio: activeToRunningTasksRatioDefault,
                     addTasksIntervalInSeconds: addTasksIntervalInSecondsDefault,
                     maxWallClockTimeInHours: maxWallClockTimeInHours,
-                };
+                } as JobManagerConfig;
```
AhmedAbdoOrtiga: same here
can you add a value for "sendNotificationTasksCount" instead of casting, and then check the value in the UTs something like: **expect(poolLoadGenerator.sendNotificationTasksCount).toEqual(x);**

packages/azure-services/src/azure-batch/pool-load-generator.spec.ts
```diff
@@ -26,7 +26,7 @@ describe(PoolLoadGenerator, () => {
                     activeToRunningTasksRatio: activeToRunningTasksRatio,
                     addTasksIntervalInSeconds: 15,
                     maxWallClockTimeInHours: 1,
-                };
+                } as JobManagerConfig;
```
AhmedAbdoOrtiga: can you add a value for "sendNotificationTasksCount" instead of casting?, and then check the value in the UTs something like: expect(poolLoadGenerator.sendNotificationTasksCount).toEqual(x);

packages/job-manager/src/batch/pool-load-generator.spec.ts
devaradhanm: nope. prefer to only set the properties that we need. this way you dont have to update this test whenever a new property is added.

packages/azure-services/src/azure-batch/pool-load-generator.spec.ts
devaradhanm: same as above

PR #661, opened 2020-03-26
fix: resize batch pools

### LINKED ISSUE #628: [Bug] Avoid using standard_d11_v2 for dev environment
## Describe the bug
Currently we have four batch pools.
2 vmss(standard_d1_v2) with 2 nodes & 1 core count - 1.65$ per day each
2 vmss(standard_d11_v2) with 3 nodes & 2 core count - 6.45$ per day each.

We don't need the higher configuration for dev environment. At least not for the legacy pool.
For the new workflow pool we can use standard_d1_v2 with 2 nodes or standard_d11_v2 with 1 node. We can start with former & then move to standard_d1_v2 if we see perf issues.

Using standard_d1_v2  with 2 nodes will bring down cost from 12.9$ to 3.3$ for these two pools per day.

## Area

- [ ] Install script
- [ ] Update script
- [ ] Service

pownkel:
#### Description of changes

- Resize all batch pools to 1 node for dev and ci environments
- Scale up url-scan-pool from 3 to 6 nodes in prod and ppe

This should bring batch costs down by more than half in dev resource groups (from ~$17 to ~$8 per day)

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #628

### DIFF
--- a/packages/resource-deployment/templates/batch-account-dev.parameters.json
+++ b/packages/resource-deployment/templates/batch-account-dev.parameters.json
@@ -2,12 +2,6 @@
     "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
     "contentVersion": "1.0.0.0",
     "parameters": {
-        "scanRequestPoolNodes": {
-            "value": "1"
-        },
-        "urlScanPoolNodes": {
-            "value": "1"
-        },
         "onDemandScanRequestPoolNodes": {
             "value": "1"
         },
--- a/packages/resource-deployment/templates/batch-account-dev.parameters.json
+++ b/packages/resource-deployment/templates/batch-account-dev.parameters.json
@@ -1,0 +1,30 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "scanRequestPoolVmSize": {
+            "value": "standard_d1_v2"
+        },
+        "scanRequestPoolNodes": {
+            "value": "1"
+        },
+        "urlScanPoolVmSize": {
+            "value": "standard_d11_v2"
+        },
+        "urlScanPoolNodes": {
+            "value": "1"
+        },
+        "onDemandScanRequestPoolVmSize": {
+            "value": "standard_d1_v2"
+        },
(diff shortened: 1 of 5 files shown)


packages/resource-deployment/templates/batch-account-prod.parameters.json
```diff
@@ -0,0 +1,18 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "scanRequestPoolNodes": {
```
lamaks: we are depreciating legacy workflow and can drop `scanRequestPoolNodes` and `urlScanPoolNodes` completely

packages/resource-deployment/templates/batch-account-dev.parameters.json
```diff
@@ -0,0 +1,18 @@
+{
+    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentParameters.json#",
+    "contentVersion": "1.0.0.0",
+    "parameters": {
+        "scanRequestPoolNodes": {
```
lamaks: we are depreciating legacy workflow and can drop `scanRequestPoolNodes` and `urlScanPoolNodes` completely

packages/resource-deployment/templates/batch-account.template.json
```diff
@@ -36,6 +36,34 @@
             "metadata": {
                 "description": "Name of the virtual network to be used for the batch pool nodes"
             }
+        },
+        "scanRequestPoolNodes": {
```
lamaks: we are depreciating legacy workflow and can drop `scanRequestPoolNodes` and `urlScanPoolNodes` completely

packages/resource-deployment/templates/batch-account-dev.parameters.json
devaradhanm: @lamaks , we can drop all the legacy stuff in another pr. 
(if we just remove it, we need to also remove the job schedule that depends on them etc)

packages/resource-deployment/templates/batch-account.template.json
pownkel: I removed the parameters for the legacy pools and hardcoded them to 1, but since removing the legacy flow is going to change more files I think that should be a seperate PR

PR #668, opened 2020-03-27
chore: Set Scan Request Sender Job Priority to Maximum

AhmedAbdoOrtiga:
#### Description of changes

- Set scan request sender job priority to 1000, while notification job is still 0, so scan request sender won't starve if too many notifications are being sent.

#### Pull request checklist

### DIFF
--- a/packages/resource-deployment/templates/on-demand-scan-req-schedule.template.json
+++ b/packages/resource-deployment/templates/on-demand-scan-req-schedule.template.json
@@ -4,7 +4,7 @@
         "recurrenceInterval": "PT2M"
     },
     "jobSpecification": {
-        "priority": 0,
+        "priority": 1000,
         "constraints": {
             "maxWallClockTime": "PT1H",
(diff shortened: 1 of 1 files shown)


packages/resource-deployment/templates/on-demand-scan-req-schedule.template.json
```diff
@@ -4,7 +4,7 @@
         "recurrenceInterval": "PT2M"
     },
     "jobSpecification": {
-        "priority": 0,
+        "priority": 1000,
```
devaradhanm: P2: i would suggest not to use the maximum value. instead use a smaller value that is greater than all other jobs we currently have, since if we want to have another job with higher priority, this will make us to reset this priority first.

AhmedAbdoOrtiga: I don't think we will have another job that has higher priority than scan request sender, and if we do that means we should assign it in another pool or at least change the pool name

PR #677, opened 2020-03-30
fix: update pool config

pownkel:
#### Description of changes

- Reduce node count in on-demand-scan-request-pool from 2 to 1 in dev environment
- Change vmSize in on-demand-scan-request-pool from standard_d1_v2 to standard_d11_v2 in prod and ppe
- Increase maxTasksPerNode in on-demand-scan-request-pool from 1 to 4 and set runExclusive to false to allow tasks to run in parallel
- Add template parameters for on-demand-url-scan-pool
- During batch account deployment, check deployed pool configurations against the parameter files and delete them if certain properties do not match (maxTasksPerNode and vmSize cannot be changed without completely deleting and recreating the pool)

#### Pull request checklist


packages/resource-deployment/scripts/batch-account-create.sh
```diff
@@ -74,6 +78,10 @@ echo "Setting up batch account $batchAccountName"
 # Configure Azure subscription account to support Batch account in user subscription mode
 . "${0%/*}/account-set-batch-app.sh"
 
+setParameterFilePath
+
+. "${0%/*}/delete-outdated-pools.sh"
```
devaradhanm: you should be running this command after disabling the schedule & waiting for noodes to be idle.

you could update the recreate pools script to do this. -https://github.com/microsoft/accessibility-insights-service/blob/master/packages/resource-deployment/scripts/recreate-vmss-for-pools.sh

checkIfVmsssAreOld could also look at the pool config & delete all pools if one of them is outdated.
And move the recreate pools script to be run before batch-account-create.sh.

devaradhanm: you can remove this deployment script change & work on it in a separate pr. Lets merge the changes for the pool config now.

PR #681, opened 2020-03-30
fix: delete outdated pools in deployment script

pownkel:
#### Description of changes

Add a deployment script to check the currently deployed batch pool config against our files and delete pools that need to be recreated. This will remove the need to delete batch pools before deploying changes to maxTasksPerNode and vmSize.

This script runs just before deploying the batch script, and has been merged with the recreate-vmss-for-pools script to check the creation date and also delete pools if vmss need to be rotated.

#### Pull request checklist


packages/resource-deployment/scripts/delete-outdated-pools.sh
```diff
@@ -0,0 +1,149 @@
...
+        return
+    fi
+
+    compareConfigFileToDeployedConfig $poolId "vmSize" "${poolPropertyNamePrefix}VmSize"
+    if [ $shouldDeletePool == "true" ]; then
+        deletePool $poolId
+        return
+    fi
+
+    compareConfigFileToDeployedConfig $poolId "maxTasksPerNode" "${poolPropertyNamePrefix}MaxTasksPerNode"
```
devaradhanm: can u also add a flag that can be passed via command line to drop nodes if the flag is set?

packages/resource-deployment/scripts/delete-outdated-pools.sh
```diff
@@ -0,0 +1,149 @@
...
+    b) batchAccountName=${OPTARG} ;;
+    p) parameterFilePath=${OPTARG} ;;
+    t) deleteTimeout=${OPTARG} ;;
+    r) resourceGroupName=${OPTARG} ;;
+    *) exitWithUsageInfo ;;
+    esac
+done
+
+# Print script usage help
+if [[ -z $batchAccountName ]] || [[ -z $parameterFilePath ]] || [[ -z $resourceGroupName ]]; then
```
devaradhanm: don't ask for inputs that we can figure out ourselves in code. call get-resource-names.sh to set the batchAccountName

packages/resource-deployment/scripts/delete-outdated-pools.sh
```diff
@@ -0,0 +1,149 @@
...
+batchAccountExists=$(az resource list --name $batchAccountName -o tsv)
+if [[ -z "$batchAccountExists" ]]; then
+    echo "batch account $batchAccountName has not yet been created."
+else
+    az batch account login --name "$batchAccountName" --resource-group "$resourceGroupName"
+
+    . "${0%/*}/process-utilities.sh"
+
+    deletePoolsIfOutdatedProcesses=(
+        "deletePoolIfOutdated \"on-demand-scan-request-pool\" \"onDemandScanRequestPool\""
```
devaradhanm: don't hardcode pool name. fetch all pools under the batch & then iterate over each pool

packages/resource-deployment/scripts/run-command-when-batch-nodes-are-idle.sh
```diff
@@ -117,7 +128,11 @@ function runCommand() {
     echo "Logging into '$batchAccountName' Azure Batch account"
     az batch account login --name "$batchAccountName" --resource-group "$resourceGroupName"
 
-    pools=$(az batch pool list --query "[].id" -o tsv)
+    if [[ -z "$poolId" ]]; then
+        pools=$(az batch pool list --query "[].id" -o tsv)
```
lamaks: does it return an array? if yes we need normalize parameter to accept array as well

packages/resource-deployment/scripts/run-command-when-batch-nodes-are-idle.sh
```diff
@@ -7,17 +7,18 @@ set -eo pipefail
 
 exitWithUsageInfo() {
     echo "
-Usage: $0 -r <resource group> -c <command to run> -n <command name>
+Usage: $0 -r <resource group> -c <command to run> -n <command name> [-p <pool id (if not specified, will wait for all pools to become idle)>]
```
lamaks: pool id or pool is list?

packages/resource-deployment/scripts/delete-outdated-pools.sh
```diff
@@ -0,0 +1,149 @@
...
+        poolExists=true
+    fi
+}
+
+deletePoolWhenNodesAreIdle() {
+    poolId=$1
+
+    command="deletePool $poolId"
+    commandName="Delete pool $poolId"
+    . "${0%/*}/run-command-when-batch-nodes-are-idle.sh"
```
devaradhanm: i would suggest to execute deletion of all pools as a single command to run-command-when-batch-nodes-are-idle, instead of calling run-command-when-batch-nodes-are-idle.sh for each pool.

packages/resource-deployment/scripts/delete-outdated-pools.sh
devaradhanm: P2: why not merge this & recreate-vmss-for-pools.sh?

packages/resource-deployment/scripts/delete-outdated-pools.sh
```diff
@@ -0,0 +1,149 @@
...
+    batchConfigPropertyName=$2
+    templateFileParameterName=$3
+
+    query="[?id=='$poolId'].$batchConfigPropertyName"
+
+    expectedValue=$(cat $parameterFilePath | jq -r ".parameters.$templateFileParameterName.value")
+    actualValue=$(az batch pool list --account-name "$batchAccountName" --query "$query" -o tsv)
+
+    if [ "$expectedValue" != "$actualValue" ]; then
+        echo "$batchConfigPropertyName for $poolId must be updated from $actualValue to $expectedValue."
```
lamaks: ... Deleting pool to perform pool update.

packages/resource-deployment/scripts/delete-outdated-pools.sh
devaradhanm: so, you condition to drop nodes would be:
is dropNodesVariable set?
is the pool config different?
are the vmss nodes in the pool are outdated (by calling 
```
az vmss list \
        --query "[?tags.BatchAccountName=='$batchAccountName' and tags.PoolName == 'pool-name1'].tags.VmssCreatedDate" \
        -o tsv)
```
?

packages/resource-deployment/scripts/run-command-when-batch-nodes-are-idle.sh
pownkel: I just changed it to accept an array of pools

packages/resource-deployment/scripts/batch-account-create.sh
```diff
@@ -51,11 +52,12 @@ function deployBatch() {
 }
 
 # Read script arguments
-while getopts ":r:t:e:" option; do
+while getopts ":r:t:e:d" option; do
     case $option in
     r) resourceGroupName=${OPTARG} ;;
     t) batchTemplateFile=${OPTARG} ;;
     e) environment=${OPTARG} ;;
+    d) dropPools=true ;;
```
lamaks, line 60: should it take param value instead?

packages/resource-deployment/scripts/delete-pools-if-needed.sh
```diff
@@ -0,0 +1,218 @@
...
+        return
+    fi
+    
+    checkPoolConfigs
+    if [[ $poolConfigOutdated == true ]]; then
+        deletePoolsWhenNodesAreIdle
+        return
+    fi
+
+    echo "Pools no do not need to be recreated."
```
devaradhanm, line 188: Pool do not

packages/resource-deployment/scripts/delete-pools-if-needed.sh
```diff
@@ -0,0 +1,218 @@
...
+}
+
+# Read script arguments
+while getopts ":r:p:d" option; do
+    case $option in
+    r) resourceGroupName=${OPTARG} ;;
+    p) parameterFilePath=${OPTARG} ;;
+    d) dropPools=true ;;
```
lamaks, line 196: should it take param value instead?

packages/resource-deployment/scripts/install-parallel.sh
```diff
@@ -106,6 +107,7 @@ while getopts ":r:s:l:e:o:p:c:t:v:k:" option; do
     t) webApiAdClientSecret=${OPTARG} ;;
     v) releaseVersion=${OPTARG} ;;
     k) enableSoftDeleteOnKeyVault=${OPTARG} ;;
+    d) dropPools=true ;;
```
lamaks, line 110: should it take param value instead?

packages/resource-deployment/scripts/delete-pools-if-needed.sh
```diff
@@ -0,0 +1,218 @@
...
+}
+
+function deletePoolsIfNeeded() {
+    az batch account login --name "$batchAccountName" --resource-group "$resourceGroupName"
+    pools=$(az batch pool list --query "[].id" -o tsv)
+    if [[ -z "$pools" ]]; then
+        return
+    fi
+
+    if [[ "$dropPools" == true ]]; then
```
devaradhanm, line 171: P2: you could do something like this:

```
if (dropPools == false) {
    checkIfVmssAreOld
    if(areVmssOld == false) {
        checkPoolConfigs
        if(poolConfigOutdated == false) {
           echo "Not crecreating pools"
           return
        }
    }
}
deletePoolsWhenNodesAreIdle


```

packages/resource-deployment/scripts/delete-pools-if-needed.sh
pownkel: Right now it's just a flag that doesn't accept parameters, so we set this variable to true if the script is run with the flag enabled. Which is preferred?

packages/resource-deployment/scripts/delete-pools-if-needed.sh
```diff
@@ -0,0 +1,218 @@
...
+function checkIfVmssAreOld {
+    areVmssOld=false
+    local hasCreatedDateTags=false
+    
+    local createdDates=$(az vmss list \
+        --query "[?tags.BatchAccountName=='$batchAccountName'].tags.VmssCreatedDate" \
+        -o tsv
+    )
+    
+    echo "VMSS created dates: $createdDates"
```
lamaks, line 35: will have an empty date if no tags, remove under `for`?

packages/resource-deployment/scripts/delete-pools-if-needed.sh
```diff
@@ -0,0 +1,218 @@
...
+        local recycleDate=$(date -d "$createdDate+$recycleVmssIntervalDays days" "+%Y-%m-%d")
+        local currentDate=$(date "+%Y-%m-%d")
+        
+        if [[ "$currentDate" > "$recycleDate" ]] || [[ "$currentDate" == "$recycleDate" ]]; then
+            echo "Found vmss with older created date $createdDate. 
+                Expected to be not older than $recycleVmssIntervalDays days.
+                Marking all vmss as old
+            "
+            areVmssOld=true
+            break
```
lamaks, line 48: do you plan to drop all vmss-s if one has expired?

pownkel: Yes, that's what we were doing before as well

packages/resource-deployment/scripts/delete-pools-if-needed.sh
lamaks: `d` is part of usage hence should follow the pattern; otherwise just have it as file level var only

lamaks: or set default if no param specified

PR #691, opened 2020-04-01
chore: Add retry logic for message creation and keep scan request if queue fails

haonliu:
Add retry logic for message creation and keep scan request if queue fails

### DIFF
--- a/packages/azure-services/src/azure-queue/queue.spec.ts
+++ b/packages/azure-services/src/azure-queue/queue.spec.ts
@@ -214,7 +214,7 @@ describe(Queue, () => {
     });
 
     describe('createMessage', () => {
-        it('creates message & queue', async () => {
+        it('creates message & queue message', async () => {
             const messageText = 'some message';
 
             setupRetryHelperMock();
--- a/packages/azure-services/src/azure-queue/queue.spec.ts
+++ b/packages/azure-services/src/azure-queue/queue.spec.ts
@@ -4,8 +4,8 @@
 import 'reflect-metadata';
 
 import { Aborter, MessageIdURL, MessagesURL, Models, QueueURL, ServiceURL } from '@azure/storage-queue';
-import { ServiceConfiguration } from 'common';
-import { IMock, Mock, MockBehavior, Times } from 'typemoq';
+import { RetryHelper, ServiceConfiguration } from 'common';
+import { IMock, It, Mock, MockBehavior, Times } from 'typemoq';
 import { MessageIdURLProvider, MessagesURLProvider, QueueServiceURLProvider, QueueURLProvider } from '../ioc-types';
 import { MockableLogger } from '../test-utilities/mockable-logger';
 import { getPromisableDynamicMock } from '../test-utilities/promisable-mock';
@@ -28,6 +28,8 @@ describe(Queue, () => {
     let deadMessagesURLMock: IMock<MessagesURL>;
     let messageIdUrlMock: IMock<MessageIdURL>;
     let serviceConfigMock: IMock<ServiceConfiguration>;
+    let retryHelperMock: IMock<RetryHelper<boolean>>;
+    const maxAttempts = 3;
 
     beforeEach(() => {
         queue = 'queue-1';
@@ -35,6 +37,7 @@ describe(Queue, () => {
         queueURLProviderMock = Mock.ofInstance((() => {}) as any);
         messagesURLProviderMock = Mock.ofInstance((() => {}) as any);
         messageIdURLProviderMock = Mock.ofInstance((() => {}) as any);
+        retryHelperMock = Mock.ofType<RetryHelper<boolean>>();
 
         serviceURLMock = Mock.ofType<ServiceURL>();
         queueURLMock = Mock.ofType<QueueURL>();
@@ -73,6 +76,9 @@ describe(Queue, () => {
             messageIdURLProviderMock.object,
             serviceConfigMock.object,
             loggerMock.object,
+            retryHelperMock.object,
+            maxAttempts,
+            0,
         );
     });
 
@@ -211,8 +217,9 @@ describe(Queue, () => {
         it('creates message & queue', async () => {
             const messageText = 'some message';
 
+            setupRetryHelperMock();
             setupQueueCreationCallWhenQueueDoesNotExist();
-            setupVerifyCallToEnqueueMessage(messagesURLMock, messageText);
+            setupVerifyCallToEnqueueMessage(messagesURLMock, messageText, { messageId: 'id' });
 
             await testSubject.createMessage(queue, messageText);
 
@@ -222,6 +229,7 @@ describe(Queue, () => {
         it('creates message succeeded when queue exists', async () => {
             const messageText = 'some message';
 
+            setupRetryHelperMock();
             setupQueueCreationCallWhenQueueExists();
             setupVerifyCallToEnqueueMessage(messagesURLMock, messageText, { messageId: 'id' });
 
@@ -233,12 +241,17 @@ describe(Queue, () => {
 
         test.each([null, { messageId: null }])('creates message failed - response = %o', async response => {
             const messageText = 'some message';
+            let caughtError: Error;
 
+            setupRetryHelperMock();
             setupQueueCreationCallWhenQueueExists();
             setupVerifyCallToEnqueueMessage(messagesURLMock, messageText, response);
 
-            const isCreated = await testSubject.createMessage(queue, messageText);
-            expect(isCreated).toBe(false);
+            await testSubject.createMessage(queue, messageText).catch(reason => {
+                caughtError = reason;
+            });
+
+            expect(caughtError.message).toBe(`Failed to enqueue the message`);
 
             verifyAll();
         });
@@ -312,6 +325,15 @@ describe(Queue, () => {
         return serverMessages.map(m => {
             return new Message(m.messageText, m.messageId, m.popReceipt);
         });
+    }
+
+    function setupRetryHelperMock(): void {
+        retryHelperMock
+            .setup(r => r.executeWithRetries(It.isAny(), It.isAny(), maxAttempts, 0))
+            .returns(async (action: () => Promise<boolean>, errorHandler: (err: Error) => Promise<void>, _: number) => {
+                return action();
+            })
+            .verifiable();
     }
 
     function verifyAll(): void {
--- a/packages/azure-services/src/azure-queue/queue.spec.ts
+++ b/packages/azure-services/src/azure-queue/queue.spec.ts
@@ -326,6 +326,8 @@ describe(Queue, () => {
         retryHelperMock
             .setup(r => r.executeWithRetries(It.isAny(), It.isAny(), maxAttempts, 0))
             .returns(async (action: () => Promise<void>, errorHandler: (err: Error) => Promise<void>, _: number) => {
+                await errorHandler(null);
+
                 return action();
             })
             .verifiable();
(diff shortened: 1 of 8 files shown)


packages/azure-services/src/azure-queue/queue.ts
```diff
@@ -17,6 +17,9 @@ export class Queue {
         @inject(iocTypeNames.MessageIdURLProvider) private readonly messageIdURLProvider: MessageIdURLProvider,
         @inject(ServiceConfiguration) private readonly serviceConfig: ServiceConfiguration,
         @inject(ContextAwareLogger) private readonly logger: ContextAwareLogger,
+        @inject(RetryHelper) private readonly retryHelper: RetryHelper<boolean>,
+        private readonly maxAttempts: number = 3,
+        private readonly millisBetweenRetries: number = 1000,
```
lamaks, line 22: `msBetweenRetries` or `retryIntervalMilliseconds`

packages/azure-services/src/azure-queue/queue.ts
```diff
@@ -71,10 +74,14 @@ export class Queue {
...
 
     public async createMessage(queue: string, message: unknown): Promise<boolean> {
-        const queueURL = await this.getQueueURL(queue);
-        await this.ensureQueueExists(queueURL);
-
-        return this.createQueueMessage(queueURL, message);
+        return this.retryHelper.executeWithRetries(
+            () => this.tryCreateMessage(queue, message),
+            async (error: Error) => {
+                return;
```
lamaks, line 80: do we need to throw?

packages/azure-services/src/azure-queue/queue.ts
```diff
@@ -84,6 +91,18 @@ export class Queue {
         return queueProperties.approximateMessagesCount;
     }
 
+    private async tryCreateMessage(queue: string, message: unknown): Promise<boolean> {
```
lamaks, line 94: this is simple create message func b/c it will throw if fail, can rename to `createMessage`

packages/azure-services/src/azure-queue/queue.ts
```diff
@@ -71,10 +74,14 @@ export class Queue {
     }
 
     public async createMessage(queue: string, message: unknown): Promise<boolean> {
-        const queueURL = await this.getQueueURL(queue);
-        await this.ensureQueueExists(queueURL);
-
-        return this.createQueueMessage(queueURL, message);
+        return this.retryHelper.executeWithRetries(
```
lamaks, line 77: did you intend  have this as `tryCreateMessage` instead? then return false on error.

packages/azure-services/src/azure-queue/queue.ts
```diff
@@ -84,6 +91,18 @@ export class Queue {
...
 
+    private async tryCreateMessage(queue: string, message: unknown): Promise<boolean> {
+        const queueURL = await this.getQueueURL(queue);
+        await this.ensureQueueExists(queueURL);
+        const enqueued = await this.createQueueMessage(queueURL, message);
+
+        if (enqueued) {
+            return true;
+        } else {
+            throw new Error(`Failed to enqueue the message`);
```
devaradhanm, line 102: add the util.inspect(message) to the error object

packages/azure-services/src/azure-queue/queue.spec.ts
```diff
@@ -233,12 +241,17 @@ describe(Queue, () => {
...
             setupQueueCreationCallWhenQueueExists();
             setupVerifyCallToEnqueueMessage(messagesURLMock, messageText, response);
 
-            const isCreated = await testSubject.createMessage(queue, messageText);
-            expect(isCreated).toBe(false);
+            await testSubject.createMessage(queue, messageText).catch(reason => {
+                caughtError = reason;
+            });
+
+            expect(caughtError.message).toBe(`Failed to enqueue the message`);
```
devaradhanm, line 255: instead you could do, 
```suggestion
await expect(testSubject.createMessage(queue, messageText)).rejects.to(new Error(`Failed to enqueue the message');
```

packages/azure-services/src/azure-queue/queue.ts
```diff
@@ -71,10 +74,14 @@ export class Queue {
...
-        await this.ensureQueueExists(queueURL);
-
-        return this.createQueueMessage(queueURL, message);
+        return this.retryHelper.executeWithRetries(
+            () => this.tryCreateMessage(queue, message),
+            async (error: Error) => {
+                return;
+            },
+            this.maxAttempts,
+            this.millisBetweenRetries,
```
devaradhanm, line 83: you don't have to expose millisBetweenRetries via contructor. Instead to exponential delay like we do in NotificationQueueMessageSender & remove the retry logic from there.

haonliu: Our retry helper does not support exponential delay right now.

packages/azure-services/src/azure-queue/queue.ts
haonliu: there is no error object here.

packages/azure-services/src/azure-queue/queue.ts
haonliu: Done.

packages/azure-services/src/azure-queue/queue.ts
devaradhanm: i meant the message, so that you know for which message the enqueuing failed.

packages/azure-services/src/azure-queue/queue.ts
devaradhanm: with this code, then you end up retrying like 6 times for notificationRequestSender

packages/web-api-scan-request-sender/src/sender/on-demand-scan-request-sender.ts
```diff
@@ -19,13 +19,19 @@ export class OnDemandScanRequestSender {
...
+                    isEnqueueSuccessful = await this.queue.createMessage(this.storageConfig.scanQueue, message);
+                }
+
+                if (isEnqueueSuccessful === true) {
                     await this.updateOnDemandPageResultDoc(resultDoc);
                 }
 
-                await this.pageScanRequestProvider.deleteRequests([page.id]);
+                if (isEnqueueSuccessful !== false) {
+                    await this.pageScanRequestProvider.deleteRequests([page.id]);
```
AhmedAbdoOrtiga, line 33: this could be in the previous if block

packages/azure-services/src/azure-queue/queue.ts
haonliu: I'm not following, the retry helper will only try up to the maxAttempts which is 3 here.

packages/azure-services/src/azure-queue/queue.ts
haonliu: Added.

packages/azure-services/src/azure-queue/queue.ts
devaradhanm: check this code - https://github.com/microsoft/accessibility-insights-service/blob/1ff2560a03e7cbdca317375b8299e4b9e50f4ea3/packages/web-api-scan-runner/src/tasks/notification-queue-message-sender.ts#L58

packages/web-api-scan-request-sender/src/sender/on-demand-scan-request-sender.ts
haonliu: Correct me if I'm wrong, if the value is undefined we still need to delete message (I think this is to delete requests that has state not equal to 'accepted'). This is the current behavior.

packages/azure-services/src/azure-queue/queue.ts
haonliu: Why not use the retry helper that we already have?

packages/azure-services/src/azure-queue/queue.ts
```diff
@@ -84,6 +93,32 @@ export class Queue {
         return queueProperties.approximateMessagesCount;
     }
 
+    public async createQueueMessage(queueURL: QueueURL, message: unknown): Promise<unknown> {
+        const messagesURL = this.getMessagesURL(queueURL);
+
+        const response = await messagesURL.enqueue(Aborter.none, JSON.stringify(message));
+        if (_.isNil(response) || _.isNil(response.messageId)) {
+            throw new Error(`Enqueue Failed with response: ${util.inspect(response)}`);
```
devaradhanm, line 101: you don't have test for when it throws. (retry failed)

packages/common/src/system/retry-helper.spec.ts
```diff
@@ -44,20 +44,21 @@ describe(RetryHelper, () => {
...
             })
-            .verifiable(Times.exactly(2));
-        errorHandlerMock.setup(e => e(testError)).verifiable(Times.once());
-        sleepFunctionMock.setup(s => s(It.isAny())).verifiable(Times.once());
+            .verifiable(Times.exactly(3));
+        errorHandlerMock.setup(e => e(testError)).verifiable(Times.exactly(2));
+        sleepFunctionMock.setup(s => s(millisBetweenRetries)).verifiable(Times.once());
+        sleepFunctionMock.setup(s => s(millisBetweenRetries * 2)).verifiable(Times.once());
 
         const result = await testSubject.executeWithRetries(actionMock.object, errorHandlerMock.object, maxAttempts, millisBetweenRetries);
```
devaradhanm, line 63: milliSecondsBetweenRetries

lamaks: millisecond is a single word, retryIntervalMilliseconds

packages/common/src/system/retry-helper.ts
```diff
@@ -29,7 +29,7 @@ export class RetryHelper<T> {
                 if (i < maxAttempts - 1) {
                     await onRetry(lastError);
                     if (millisBetweenRetries > 0) {
-                        await this.sleepFunction(millisBetweenRetries);
+                        await this.sleepFunction(millisBetweenRetries * (i + 1));
```
lamaks, line 32: retryIntervalMilliseconds

packages/web-api-scan-request-sender/src/sender/on-demand-scan-request-sender.ts
```diff
@@ -19,13 +19,20 @@ export class OnDemandScanRequestSender {
...
+                    isEnqueueSuccessful = await this.queue.createMessage(this.storageConfig.scanQueue, message);
+                }
+
+                if (isEnqueueSuccessful === true) {
                     await this.updateOnDemandPageResultDoc(resultDoc);
                 }
 
-                await this.pageScanRequestProvider.deleteRequests([page.id]);
+                if (isEnqueueSuccessful !== false) {
+                    // only when message queue failed we keep the request
```
lamaks, line 33: we delete request but comment says different

packages/common/src/system/retry-helper.spec.ts
```diff
@@ -15,7 +15,7 @@ describe(RetryHelper, () => {
     let sleepFunctionMock: IMock<(millis: number) => Promise<void>>;
     const maxAttempts = 3;
     const returnValue = 42;
-    const millisBetweenRetries = 10;
+    const milliSecondsBetweenRetries = 10;
```
lamaks, line 18: retryIntervalMilliseconds

packages/web-api-scan-request-sender/src/sender/on-demand-scan-request-sender.ts
```diff
@@ -19,13 +19,20 @@ export class OnDemandScanRequestSender {
...
+                    await this.pageScanRequestProvider.deleteRequests([page.id]);
```
devaradhanm, line 33: why not move this code to the above if block after the below line ?
                    await this.updateOnDemandPageResultDoc(resultDoc);

haonliu: Correct me if I’m wrong, if it has other states like queued we want to delete it as well. We have a test case for that.

PR #700, opened 2020-04-02
chore: remove jest-slow-test-reporter

devaradhanm:
#### Description of changes
We added this to make it easy to identify the slow running tests to fix https://github.com/microsoft/accessibility-insights-service/issues/601

Now except for the processWebRequest (which is like integration test, all other tests take < 100ms)

#### Pull request checklist

### DIFF
--- a/jest.config.base.js
+++ b/jest.config.base.js
@@ -53,14 +53,6 @@ module.exports = {
                 outputPath: './test-results/unit/html-reporter/report.html',
             },
         ],
-        // [
-        //     'jest-slow-test-reporter',
-        //     {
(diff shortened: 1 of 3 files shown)


jest.config.base.js
```diff
@@ -53,14 +53,14 @@ module.exports = {
...
-        //     },
-        // ],
+        [
+            'jest-slow-test-reporter',
+            {
+                numTests: 20,
+                warnOnSlowerThan: 100,
+                color: true,
+            },
+        ],
```
pownkel, line 63: If we're removing this why is it being uncommented?

PR #720, opened 2020-04-13
fix: check successful response code instead of 200 for notification

devaradhanm:
#### Description of changes
This PR extends considering all result codes between 200-299 as successful instead of just 200.

#### Pull request checklist


packages/azure-services/src/storage/client.ts
```diff
@@ -1,27 +1,31 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
```
devaradhanm, line 2: this sfile should be in common. but, not moving as part of this pr.

PR #737, opened 2020-04-20
fix: dropPools environment variable

pownkel:
#### Description of changes

Don't overwrite dropPools if it wasn't passed in as an environment variable

#### Pull request checklist


packages/resource-deployment/scripts/install-parallel.sh
```diff
@@ -25,7 +25,7 @@ export releaseVersion
 export templatesFolder="${0%/*}/../templates/"
 export apiTemplates="$templatesFolder"rest-api-templates
 export enableSoftDeleteOnKeyVault=true
-export dropPools=false
+export dropPools
```
devaradhanm, line 28: you can have the default value for this here to be false.

PR #739, opened 2020-04-21
fix: Set timeout when sending the notification request

AhmedAbdoOrtiga:
#### Description of changes
- If we don’t set time out, time out might take too much time especially if we keep retrying, so the task itself times out before the post request does, so it exists without updating the notification state.

#### Pull request checklist

### DIFF
--- a/packages/web-api-send-notification-runner/src/tasks/notification-sender-web-api-client.spec.ts
+++ b/packages/web-api-send-notification-runner/src/tasks/notification-sender-web-api-client.spec.ts
@@ -69,7 +69,7 @@ describe(NotificationSenderWebAPIClient, () => {
         const requestBody = { scanId: scanId, runStatus: runStatus, scanStatus: scanStatus };
         const options = {
             body: requestBody,
-            timeout: 5000,
+            timeout: 30000,
         };
 
         postMock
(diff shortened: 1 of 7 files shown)


packages/web-api-send-notification-runner/src/tasks/notification-sender-web-api-client.spec.ts
```diff
@@ -69,6 +69,7 @@ describe(NotificationSenderWebAPIClient, () => {
         const requestBody = { scanId: scanId, runStatus: runStatus, scanStatus: scanStatus };
         const options = {
             body: requestBody,
+            timeout: 5000,
```
devaradhanm, line 72: why do we need this? i don't think it will be stuck if the url doesnt respond. it should have a default timeout. what is the default timeout?

AhmedAbdoOrtiga: since we use npm request package, so as mentioned in the timeouts section here:
https://www.npmjs.com/package/request
"Most requests to external servers should have a timeout attached, in case the server is not responding in a timely manner. Without a timeout, your code may have a socket open/consume resources for minutes or more."

PR #743, opened 2020-04-21
chore: set scan state to null if run fails

AhmedAbdoOrtiga:
#### Description of changes

- If the run failed for any reason, set the scan state to null in the notification request body.

#### Pull request checklist

### DIFF
--- a/packages/web-api-scan-runner/src/runner/runner.ts
+++ b/packages/web-api-scan-runner/src/runner/runner.ts
@@ -58,9 +58,7 @@ export class Runner {
                 timestamp: new Date().toJSON(),
                 error: null,
             },
-            scanResult: {
-                state: 'pending',
-            },
+            scanResult: null,
             reports: null,
         };
 
--- a/packages/web-api-scan-runner/src/runner/runner.ts
+++ b/packages/web-api-scan-runner/src/runner/runner.ts
@@ -58,7 +58,9 @@ export class Runner {
                 timestamp: new Date().toJSON(),
                 error: null,
             },
-            scanResult: null,
+            scanResult: {
+                state: 'pending',
+            },
             reports: null,
         };
 
--- a/packages/web-api-send-notification-runner/src/tasks/notification-sender-web-api-client.spec.ts
+++ b/packages/web-api-send-notification-runner/src/tasks/notification-sender-web-api-client.spec.ts
@@ -39,7 +39,7 @@ describe(NotificationSenderWebAPIClient, () => {
     describe('verify default options', () => {
         test.each([true, false])('verifies when throwOnFailure is %o', (throwOnFailure: boolean) => {
             // tslint:disable-next-line: no-empty
-            const defaultsMock = Mock.ofInstance((options: requestPromise.RequestPromiseOptions): any => {});
+            const defaultsMock = Mock.ofInstance((options: requestPromise.RequestPromiseOptions): any => { });
             requestStub.defaults = defaultsMock.object;
 
             defaultsMock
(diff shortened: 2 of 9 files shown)


packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -58,7 +58,9 @@ export class Runner {
                 timestamp: new Date().toJSON(),
                 error: null,
             },
-            scanResult: null,
+            scanResult: {
```
devaradhanm, line 61: why can't we just set scanResult only when available instead of setting it to pending?
scanResult can instead be set to undefined.

Also, Ashwin's comment was on scan notification payload containing empty string. can we fix that?

AhmedAbdoOrtiga: - because in the beginning of the task, we start scanning, means scanning is pending, it might fail or pass at the end of it, but if it's null, it means it's we don't know what was the state of the scan, also the 'pending' value was put for that but was never used!

- if we set value to pending, then we will never send empty scanState in case of run failure which solves Ashwin's comment

devaradhanm: scanStatus should only be set when runStatus is completed. We use the runStatus for progress & scanStatus for the scan outcome.

packages/web-api-send-notification-runner/src/tasks/notification-sender-web-api-client.ts
```diff
@@ -34,7 +34,7 @@ export class NotificationSenderWebAPIClient {
         const requestBody = {
             scanId: notificationSenderConfigData.scanId,
             runStatus: notificationSenderConfigData.runStatus,
-            scanStatus: notificationSenderConfigData.scanStatus,
+            scanStatus: notificationSenderConfigData.runStatus === 'failed' ? null : notificationSenderConfigData.scanStatus,
```
devaradhanm, line 37: use isEmpty(notificationSenderConfigData.scanStatus) ? undefined : notificationSenderConfigData.scanStatus

Also, in runner set the scanStatus to undefined instead of null.

packages/web-api-send-notification-runner/src/tasks/notification-sender-web-api-client.ts
```diff
@@ -34,7 +35,7 @@ export class NotificationSenderWebAPIClient {
         const requestBody = {
             scanId: notificationSenderConfigData.scanId,
             runStatus: notificationSenderConfigData.runStatus,
-            scanStatus: notificationSenderConfigData.scanStatus,
+            scanStatus: isEmpty(notificationSenderConfigData.scanStatus) ? undefined : notificationSenderConfigData.scanStatus,
```
devaradhanm, line 38: missing tests

packages/web-api-send-notification-runner/src/tasks/notification-sender-web-api-client.spec.ts
```diff
@@ -53,33 +53,34 @@ describe(NotificationSenderWebAPIClient, () => {
         });
     });
 
-    it('postReplyUrl', async () => {
-        const scanNotifyUrl = 'scanNotifyUrl';
-        const scanId = 'scanId';
-        const scanStatus = 'pass';
-        const runStatus = 'completed';
+    describe('postReplyUrl', () => {
+        test.each(['pass', undefined])('verifies when scanState is %o', async (scanStatus: any) => {
```
devaradhanm, line 57: add empty string in to this test case. (the case u are trying to fix :))

PR #768, opened 2020-05-01
chore (scan-file-in-cli): Refactor runner command

AhmedAbdoOrtiga:
#### Description of changes

- Add another parameters in yargs called filePath
- Only one parameter should be passed url or filePath
- Refactor runner command to be an interface that has 2 implementations:
  - URL runner command which is the current implementation for single url.
  - File runner command which will read the file line by line and scan each **[to be implemented].**
- Moved writing the report content to disk to a single responsibility class called ReportDiskWriter.
  - **Sill using sync method as it is, I tried to use async but mocking might be exhausting to be done in this PR, so will look at it later**

#### Pull request checklist


packages/cli/src/cli.ts
```diff
@@ -11,11 +12,19 @@ import { setupCliContainer } from './setup-cli-container';
...
+        .usage('Usage: $0 --url <url> --filePath <filePath> --output <directoryPath>')
         .options({
-            url: { type: 'string', describe: 'url to scan for accessibility', demandOption: true },
+            url: { type: 'string', describe: 'url to scan for accessibility' },
+            filePath: { type: 'string', describe: 'file path that contians multiple urls to scan' },
             output: { type: 'string', describe: 'output directory' },
         })
+        .check((args) => {
+            if ((isEmpty(args.url) && isEmpty(args.filePath)) || (!isEmpty(args.url) && !isEmpty(args.filePath))) {
+                throw new Error('You should input either url or filePath (only one)!');
```
lamaks, line 23: Will user see error stack when error is thrown? May be user message can be like `You should provide either url or filePath parameter only`

packages/cli/src/report/report-disk-writer.ts
```diff
@@ -0,0 +1,52 @@
...
+        if (isEmpty(directory)) {
+            // tslint:disable-next-line: no-parameter-reassignment
+            directory = '.';
+        }
+
+        const reportFileName = `${directory}/${filenamify(fileName, { replacement: '_' })}.${format}`;
+
+        if (!this.fileSystemObj.existsSync(directory)) {
+            console.log('output directory does not exists.');
+            console.log(`creating output directory - ${directory}`);
```
lamaks, line 22: probably we do not need this console messages

packages/cli/src/report/report-disk-writer.ts
```diff
@@ -0,0 +1,52 @@
...
+            this.fileSystemObj.mkdirSync(directory);
+        }
+
+        this.fileSystemObj.writeFileSync(reportFileName, content);
+
+        // this.fileSystemObj.exists(directory, async (exists) => {
```
lamaks, line 28: remove unused code

packages/cli/src/report/report-formats.ts
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+export enum ReportFormats {
```
lamaks, line 3: we can use type instead of enum `export declare type ReportFormats = 'html' | 'json';`

packages/cli/src/runner/file-command-runner.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    public async runCommand(scanArguments: ScanArguments): Promise<void> {
+        const spinner = new Spinner('Running scanner...');
+        let axeResults: AxeScanResults;
+
+        try {
+            spinner.start();
+
+            axeResults = await this.scanner.scan(scanArguments.url);
+        } finally {
```
lamaks, line 30: do we need to handle error?

packages/cli/src/runner/url-command-runner.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    public async runCommand(scanArguments: ScanArguments): Promise<void> {
+        const spinner = new Spinner('Running scanner...');
+        let axeResults: AxeScanResults;
+
+        try {
+            spinner.start();
+
+            axeResults = await this.scanner.scan(scanArguments.url);
+        } finally {
```
lamaks, line 30: do we need to handle error?

packages/cli/src/cli.ts
AhmedAbdoOrtiga: he'll see the error message, because we have a catch block that prints it then exits

packages/cli/src/report/report-disk-writer.ts
AhmedAbdoOrtiga: I wanted to keep this code to switch to async file operations in another pr, but it's ok I can add it later

packages/cli/src/report/report-disk-writer.ts
AhmedAbdoOrtiga: I think it's good to keep it for at least debugging purposes

packages/cli/src/runner/file-command-runner.ts
AhmedAbdoOrtiga: I don't think it will add any benefit and that's why it wasn't added before, we will print the error and exit which is done by default

packages/cli/src/runner/url-command-runner.ts
AhmedAbdoOrtiga: I don't think it will add any benefit and that's why it wasn't added before, we will print the error and exit which is done by default

packages/cli/src/cli-entry-point.ts
```diff
@@ -2,14 +2,27 @@
...
 export class CliEntryPoint {
     constructor(private readonly container: Container) {}
 
     public async runScan(scanArguments: ScanArguments): Promise<void> {
-        const commandRunner = this.container.get(CommandRunner);
+        const commandRunner = this.commandRunnerFactory(scanArguments);
         await commandRunner.runCommand(scanArguments);
     }
+
+    private commandRunnerFactory(scanArguments: ScanArguments): CommandRunner {
```
devaradhanm, line 19: method names should not be noun. rename to getCommandRunner

packages/cli/src/cli.ts
```diff
@@ -11,11 +12,19 @@ import { setupCliContainer } from './setup-cli-container';
 
 (async () => {
     const scanArguments = (yargs
-        .usage('Usage: $0 --url <url> --output <directoryPath>')
+        .usage('Usage: $0 --url <url> --filePath <filePath> --output <directoryPath>')
```
devaradhanm, line 15: rename to inputFilePath or inputFile

devaradhanm: else it is confusing to know if the filePath is for output or input

packages/cli/src/cli.ts
```diff
@@ -11,11 +12,19 @@ import { setupCliContainer } from './setup-cli-container';
...
         .options({
-            url: { type: 'string', describe: 'url to scan for accessibility', demandOption: true },
+            url: { type: 'string', describe: 'url to scan for accessibility' },
+            filePath: { type: 'string', describe: 'file path that contians multiple urls to scan' },
```
devaradhanm, line 18: file path that contians multiple urls separated by newline to scan

devaradhanm, line 18: contians -> typo

packages/cli/src/report/report-disk-writer.ts
```diff
@@ -0,0 +1,29 @@
...
+export class ReportDiskWriter {
+    constructor(private readonly fileSystemObj: typeof fs = fs) {}
+
+    public writeToDirectory(directory: string, fileName: string, format: ReportFormats, content: string): void {
+        if (isEmpty(directory)) {
+            // tslint:disable-next-line: no-parameter-reassignment
+            directory = '.';
+        }
+
+        const reportFileName = `${directory}/${filenamify(fileName, { replacement: '_' })}.${format}`;
```
devaradhanm, line 19: we may not be able to use url value as file name since you may get path too long exceptions. (but this doesn't have to be solved in this pr)

packages/cli/src/runner/file-command-runner.ts
```diff
@@ -0,0 +1,40 @@
...
+    ) {}
+
+    public async runCommand(scanArguments: ScanArguments): Promise<void> {
+        const spinner = new Spinner('Running scanner...');
+        let axeResults: AxeScanResults;
+
+        try {
+            spinner.start();
+
+            axeResults = await this.scanner.scan(scanArguments.url);
```
devaradhanm, line 28: this code is same as UrlCommandRunner. Shouldn't this read the inputFilePath & then iterate over the urls to scan?

devaradhanm: Or are you planning to implement this in next PR?

AhmedAbdoOrtiga: yeah I mentioned in the PR description that this will be done in another PR, I was just refactoring and getting things ready for next PR

packages/cli/src/runner/url-command-runner.ts
devaradhanm: but you are handling error in FileCommandRunner?

packages/cli/src/report/report-disk-writer.ts
AhmedAbdoOrtiga: okies, we can look at this in another PR

packages/cli/src/cli.ts
```diff
@@ -11,11 +12,19 @@ import { setupCliContainer } from './setup-cli-container';
 
 (async () => {
     const scanArguments = (yargs
-        .usage('Usage: $0 --url <url> --output <directoryPath>')
+        .usage('Usage: $0 --url <url> --inputFile <inputFile> --output <directoryPath>')
         .options({
-            url: { type: 'string', describe: 'url to scan for accessibility', demandOption: true },
+            url: { type: 'string', describe: 'url to scan for accessibility' },
+            inputFile: { type: 'string', describe: 'ile path that contains multiple urls separated by newline to scan' },
```
devaradhanm, line 18: fyi: when parsing the urls from the file, you might need to handle \n, \r, \r\n as new line character

packages/cli/src/cli-entry-point.spec.ts
```diff
@@ -5,29 +5,48 @@ import 'reflect-metadata';
...
+            // tslint:disable-next-line: no-floating-promises
+            await testSubject.runScan(testInput);
+            expect(runCommand).toBeCalled();
+        });
+
+        it('returns null', async () => {
+            const testInput: ScanArguments = { output: '/users/xyz' };
+            // tslint:disable-next-line: no-floating-promises
+            await testSubject.runScan(testInput).catch((error) => {
+                expect((error as Error).message).toEqual('You should provide either url or inputFile parameter only!');
```
devaradhanm, line 48: have you verified it will fail if the error message doesnt match "You should provide either url or inputFile parameter only!" ?

devaradhanm: you could do something like this https://jestjs.io/docs/en/expect#rejects

await expect(Promise.reject(new Error('octopus'))).rejects.toThrow('octopus');

packages/cli/src/cli.ts
AhmedAbdoOrtiga: cool, good to know.

packages/cli/src/runner/file-command-runner.ts
```diff
@@ -0,0 +1,40 @@
...
+        } catch (error) {
+            console.log('Exception thrown while running scanner: ', error);
+            process.exit(1);
```
devaradhanm, line 31: catch code doesnt have any unit tests

packages/cli/src/cli-entry-point.spec.ts
AhmedAbdoOrtiga: yes I did, if fails if the message doesn't match
actually the pattern you mentioned doesn't work well, it always pass if the message doesn't match

packages/cli/src/runner/file-command-runner.ts
AhmedAbdoOrtiga: I forgot to remove this code, I replied to Maxim about it in another comment

packages/cli/src/cli.ts
devaradhanm, line 18: typo. file not ile.

packages/cli/src/cli-entry-point.spec.ts
```diff
@@ -5,29 +5,46 @@ import 'reflect-metadata';
...
+            containerMock.setup((cm) => cm.get(FileCommandRunner)).returns(() => fileCommandRunnerMock.object);
+            const runCommand = jest.spyOn(fileCommandRunnerMock.object, 'runCommand').mockImplementationOnce(async () => Promise.resolve());
+            // tslint:disable-next-line: no-floating-promises
+            await testSubject.runScan(testInput);
+            expect(runCommand).toBeCalled();
+        });
+
+        it('returns null', async () => {
+            const testInput: ScanArguments = { output: '/users/xyz' };
+            // tslint:disable-next-line: no-floating-promises
```
devaradhanm, line 46: shoulld you still disable  no-floating-promises?

packages/cli/src/runner/url-command-runner.ts
AhmedAbdoOrtiga: nope, forgot to remove it

PR #773, opened 2020-05-04
chore: update build/deploy VM images to ubuntu 18.04

dbjorge:
#### Description of changes

Updates the batch pool and devops VM images to use latest ubuntu LTS (18.04), from 16.04.

This removes the "npm" demand from our pool specification; it isn't necessary since we install node explicitly anyway in all of our builds, and it causes a build failure ([example](https://dev.azure.com/accessibility-insights/Accessibility%20Insights%20Service/_build/results?buildId=8738&view=results)) where the builds to try to use a non-existent "Hosted Ubuntu 18.04" agent pool instead of the more typical "Azure Pipelines" default agent pool that all our other builds use.

I considered separating these to 2 PRs, but thought it was probably best to pin what we test against and what we deploy against to one another.

I'm not familiar with how the batch pools work, particularly, whether this is a change that might orphan old pool VMs somehow; would like a review from someone more familiar with how batch pool machine deployment works

#### Pull request checklist

### DIFF
--- a/packages/resource-deployment/templates/batch-account.template.json
+++ b/packages/resource-deployment/templates/batch-account.template.json
@@ -118,10 +118,10 @@
                         "imageReference": {
                             "publisher": "microsoft-azure-batch",
                             "offer": "ubuntu-server-container",
-                            "sku": "16-04-lts",
+                            "sku": "18-04-lts",
                             "version": "latest"
                         },
-                        "nodeAgentSkuId": "batch.node.ubuntu 16.04"
+                        "nodeAgentSkuId": "batch.node.ubuntu 18.04"
                     }
                 },
                 "networkConfiguration": {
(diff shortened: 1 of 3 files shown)


packages/resource-deployment/templates/batch-account.template.json
```diff
@@ -118,10 +118,10 @@
                         "imageReference": {
                             "publisher": "microsoft-azure-batch",
                             "offer": "ubuntu-server-container",
-                            "sku": "16-04-lts",
+                            "sku": "18-04-lts",
```
devaradhanm, line 121: @dbjorge  have you validated this change by deploying it using https://dev.azure.com/accessibility-insights-private/Accessibility%20Insights%20(private)/_release?_a=releases&view=mine&definitionId=12

Also, since you are changing the image, we need to drop nodes & create again. (this can be achieved by setting the release environment variable dropPools=true (Else the deployment will fail)

dbjorge:
Motivation was just to keep dependencies up to date and consistent across repos. Since it sounds like there's a known incompatibility with 18 and the DevTest deployment failed citing an invalid batch offer on the updated image name, closing/abandoning this PR for now.

PR #774, opened 2020-05-04
chore (scan-file-in-cli): run ai scan on a file that has multiple urls separated by a new line

AhmedAbdoOrtiga:
#### Description of changes

- Read input file line by line.
- Scan each url if it wasn't scanned before.
- Save result and update summary.

#### Pull request checklist


packages/cli/src/report/report-disk-writer.ts
```diff
@@ -10,13 +10,19 @@ import { ReportFormats } from './report-formats';
...
             directory = '.';
         }
 
-        const reportFileName = `${directory}/${filenamify(fileName, { replacement: '_' })}.${format}`;
+        let reportFileName;
+
+        try {
+            reportFileName = `${directory}/${filenamify(fileName, { replacement: '_' })}.${format}`;
+        } catch {
+            reportFileName = `${directory}/${fileName}.${format}`;
```
devaradhanm, line 24: why do we need this try catch

packages/cli/src/runner/file-command-runner.ts
```diff
@@ -2,36 +2,95 @@
...
 
-            axeResults = await this.scanner.scan(scanArguments.url);
+            lines.forEach(async (line) => {
+                // tslint:disable-next-line: no-parameter-reassignment
+                line = line.trim();
+                if (!this.lodashObj.isEmpty(line) && !this.uniqueUrls.has(line)) {
+                    this.uniqueUrls.add(line);
+                    promises.push(
+                        this.scanURL(line).then((reportContent) => {
+                            const reportName = this.reportDiskWriter.writeToDirectory(scanArguments.output, line, 'html', reportContent);
```
devaradhanm, line 54: even though you are trying to run this in parallel, it will be running in serial, since writeToDirectory is sync. can you change that?

packages/cli/src/runner/file-command-runner.ts
```diff
@@ -2,36 +2,95 @@
...
+    private readonly uniqueUrls = new Set();
+
     constructor(
         @inject(AIScanner) private readonly scanner: AIScanner,
         @inject(ReportGenerator) private readonly reportGenerator: ReportGenerator,
         @inject(ReportDiskWriter) private readonly reportDiskWriter: ReportDiskWriter,
+        @inject(JsonSummaryReportGenerator) private readonly jsonSummaryReportGenerator: JsonSummaryReportGenerator,
+        @inject(ConsoleSummaryReportGenerator) private readonly consoleSummaryReportGenerator: ConsoleSummaryReportGenerator,
+        private readonly fileSystemObj: typeof fs = fs,
+        private readonly lodashObj: typeof lodash = lodash,
```
lamaks, line 35: do we need this property for lodash? The import only should be sufficient

packages/cli/src/runner/file-command-runner.ts
```diff
@@ -2,36 +2,95 @@
...
+                this.jsonSummaryReportGenerator.generateReport(this.summaryReportData),
+            );
+            console.log(`ViolationCountByRuleMap summary was saved in file ${jsonSummryReportName}`);
+        });
+    }
+
+    private async scanURL(url: string): Promise<string> {
+        let axeResults: AxeScanResults;
+
+        axeResults = await this.lodashObj.cloneDeep(this.scanner).scan(url);
```
lamaks, line 80: this is not kind of object we usually clone. The scanner is setup as singleton hence we need to change to factory pattern instead

https://github.com/microsoft/accessibility-insights-service/blob/2c2391db6dd40ed17224decf35e6a2c25d776ad2/packages/cli/src/setup-cli-container.ts#L18

packages/cli/src/report/report-disk-writer.ts
AhmedAbdoOrtiga: if the file name has space, it throws exception!

packages/cli/src/runner/file-command-runner.ts
AhmedAbdoOrtiga: I'm trying to scan in parallel since scan is a promise, also each writeTo Disk should be after its scan

packages/cli/src/runner/file-command-runner.ts
AhmedAbdoOrtiga: we use this pattern in all our files so we can mock its functions in the uts

packages/cli/src/runner/file-command-runner.ts
AhmedAbdoOrtiga: the issue isn't the pattern, the issue is the way we inject our dependency, we need a new scanner for each can, but we inject the scanner in the constructor so it will be the same every for all scans, which throws exception since scan runs parallel

packages/cli/src/runner/file-command-runner.ts
devaradhanm: if the code inside the async function is running synchronously, then trying to run multiple invocations in parallel wont execute them in paralle. You can test it yourselves by doing some logging.

packages/cli/src/runner/file-command-runner.ts
lamaks: Why we need to mock lodash functions? I do not see any code in project where we do mock it. We can mock fs that is okay, but mock lodash seems odd.

If we move to factory pattern for `AIScanner` instance creation we do not need to mock lodash.

packages/cli/src/runner/file-command-runner.ts
devaradhanm: you should never clone object that has functions. cloning should be done only on data object.
In this case, if you want to create multiple scanner:
1. ) you should limit the parallel scanners that are being executed. You can use https://www.npmjs.com/package/p-limit
2.) In this case you are trying to have multiple instances of scanner. I would suggest refactoring Scanner & make it stateless. (don't take page in constructor.) And also refactor Page class. (Page should not launch or close browser.)


How about something like this.:

```
const browser = webdriver.launch();

// in async with max parallel execuition of lets say 5)
urls.forEach(url => {
     await this.scanner.scan(url, browser);
})


Scanner.scan = (url: string, browser: Puppeteer.Browser) => {
    const page = this.pageFactory.create(url, browser);
    await page.create(url, browser);
    await page.scanForA11yIssues();
   ....
}



export class PageFactory {
    public create(): Page {
        return new Page(....);
    }
}
```

lamaks: We can inject factory in a constructor to use it for `AIScanner` instance creation like this for example:

https://github.com/microsoft/accessibility-insights-service/blob/2920dcab1e913e87651cad94bf9ad2f50d922314/packages/scanner/src/page.ts#L19

packages/cli/src/report/report-disk-writer.ts
lamaks: May be we need to replace it to with dash instead

AhmedAbdoOrtiga:
> We need to defer for parallel scan implementation till we integrate the crawler solution. This will save the effort.

I deferred it but It won't save any effort, the scan call is already asynchronous call, and to defer we have to make it synchronous which need more effort as done in the code

packages/cli/src/runner/file-command-runner.ts
AhmedAbdoOrtiga: again the scan used to be async so the code will be async, if you test (before the last commit), you can see that a new scan used to start before the first finishes

packages/cli/src/runner/file-command-runner.ts
AhmedAbdoOrtiga: fs is mocked in service-configuration, axe-puppeteer, etc, isn't this the same pattern?, anyway that was removed

packages/cli/src/report/report-disk-writer.ts
AhmedAbdoOrtiga: why?

packages/cli/src/runner/file-command-runner.ts
AhmedAbdoOrtiga: no cloning anymore, updated!

packages/cli/src/report/report-disk-writer.ts
lamaks: We are using a dedicated package to handle the conversion of a string to a file name and it clearly not working  as it should be. Either the package should convert correctly w/o throwing an error (we may need to adjust function call maybe?) or look for other package/solution. Error on a space is quite odd.

AhmedAbdoOrtiga: yeah the issue isn't the underscore, and failing if a url has a space is expected, because we use 'filenamify-url' which expect url, and url shouldn't has a space, it should be encoded like '%20' or something, so I'll use 'filenamify-url' and if it fails, I'll use 'filenamify' which expect just a string

PR #787, opened 2020-05-05
Limit feed trigger to batch of 10 documents

lamaks:
#### Description of changes

#### Pull request checklist

### DIFF
--- a/packages/web-workers/scan-batch-requests-feed-func/function.json
+++ b/packages/web-workers/scan-batch-requests-feed-func/function.json
@@ -9,7 +9,8 @@
             "databaseName": "onDemandScanner",
             "collectionName": "scanBatchRequests",
             "createLeaseCollectionIfNotExists": true,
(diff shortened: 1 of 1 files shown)


packages/web-workers/scan-batch-requests-feed-func/function.json
```diff
@@ -9,7 +9,8 @@
             "databaseName": "onDemandScanner",
             "collectionName": "scanBatchRequests",
             "createLeaseCollectionIfNotExists": true,
-            "startFromBeginning": true
+            "startFromBeginning": true,
```
devaradhanm, line 12: do we still need startFromBeginning? what happens if there are multiple function instances? will it make sure it doesnt send duplicate records?

lamaks: I will keep startFromBeginning It will not break but allow to re-process by trigger

PR #794, opened 2020-05-07
chore: adding support to generate summary html

devaradhanm:
#### Description of changes

This will generate a basic summary html file. Look at the jest snapshot file to get  idea on how it will look.

Refactored file-command-runner & added support to capture unscannable urls
(unscannable urls list is not surfaced anywhere in the report yet)

Refactored unit tests for file-command-runner for better coverage

scan

image

image

image

image

#### Pull request checklist

### DIFF
--- a/packages/cli/src/report/report-disk-writer.spec.ts
+++ b/packages/cli/src/report/report-disk-writer.spec.ts
@@ -23,14 +23,15 @@ describe('ReportDiskWriter', () => {
         const format = 'html';
         const content = 'content';
 
-        const reportFileName = `${directory}/${filenamify(fileName, { replacement: '_' })}.${format}`;
+        const reportFileName = `${filenamify(fileName, { replacement: '_' })}.${format}`;
+        const reportFilePath = `${directory}/${reportFileName}`;
 
         fsMock
             .setup((fsm) => fsm.existsSync('.'))
             .returns(() => true)
             .verifiable(Times.once());
         fsMock
-            .setup((fsm) => fsm.writeFileSync(reportFileName, content))
+            .setup((fsm) => fsm.writeFileSync(reportFilePath, content))
             .returns(() => {})
             .verifiable(Times.once());
         fsMock
@@ -38,7 +39,7 @@ describe('ReportDiskWriter', () => {
             .returns(() => {})
             .verifiable(Times.never());
 
-        testSubject.writeToDirectory(directory, fileName, format, content);
+        expect(testSubject.writeToDirectory(directory, fileName, format, content)).toEqual(reportFileName);
 
         fsMock.verifyAll();
     });
@@ -49,14 +50,14 @@ describe('ReportDiskWriter', () => {
         const format = 'json';
         const content = 'content';
 
-        const reportFileName = `${directory}/${fileName}.${format}`;
-
+        const reportFileName = `${fileName}.${format}`;
+        const reportFilePath = `${directory}/${reportFileName}`;
         fsMock
             .setup((fsm) => fsm.existsSync('.'))
             .returns(() => false)
             .verifiable(Times.once());
         fsMock
-            .setup((fsm) => fsm.writeFileSync(reportFileName, content))
+            .setup((fsm) => fsm.writeFileSync(reportFilePath, content))
             .returns(() => {})
             .verifiable(Times.once());
         fsMock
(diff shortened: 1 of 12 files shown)


packages/cli/src/report/report-disk-writer.spec.ts
```diff
@@ -49,22 +50,22 @@ describe('ReportDiskWriter', () => {
         const format = 'json';
         const content = 'content';
 
-        const reportFileName = `${directory}/${fileName}.${format}`;
-
+        const reportFileName = `${fileName}.${format}`;
```
AhmedAbdoOrtiga, line 53: this should be like report-disk.writer.ts, something like:

`import * as filenamify from 'filenamify';
import * as filenamifyUrl from 'filenamify-url';`
`${filenamify(fileName, { replacement: '_' })}.${format}`;
otherwise the test would fail if there's any special charterer

packages/cli/src/runner/file-command-runner.ts
```diff
@@ -30,67 +33,83 @@ export class FileCommandRunner implements CommandRunner {
...
         @inject(ReportDiskWriter) private readonly reportDiskWriter: ReportDiskWriter,
         @inject(JsonSummaryReportGenerator) private readonly jsonSummaryReportGenerator: JsonSummaryReportGenerator,
+        @inject(HtmlSummaryReportGenerator) private readonly htmlSummaryReportGenerator: HtmlSummaryReportGenerator,
         @inject(ConsoleSummaryReportGenerator) private readonly consoleSummaryReportGenerator: ConsoleSummaryReportGenerator,
         private readonly fileSystemObj: typeof fs = fs,
     ) {}
 
     public async runCommand(scanArguments: ScanArguments): Promise<void> {
-        const spinner = new Spinner(`Running scanner...\n`);
+        const spinner = new Spinner(`Running scanner... %s \t`);
```
AhmedAbdoOrtiga, line 42: Running scanner is so confusing now, it prints too many lines and after the scan is done, it looks weird if the user would look at it

packages/cli/src/runner/file-command-runner.ts
```diff
@@ -30,67 +33,83 @@ export class FileCommandRunner implements CommandRunner {
...
-    private processURLScanResult(axeResults: AxeScanResults): void {
-        axeResults.results.violations?.forEach((violation) => {
-            this.summaryReportData.violationCountByRuleMap[violation.id] =
-                // tslint:disable-next-line: strict-boolean-expressions
-                (this.summaryReportData.violationCountByRuleMap[violation.id]
-                    ? this.summaryReportData.violationCountByRuleMap[violation.id]
-                    : 0) + violation.nodes.length;
-        });
+    private processURLScanResult(url: string, reportName: string, axeResults: AxeScanResults): void {
+        if (axeResults.results.violations?.length > 0) {
```
AhmedAbdoOrtiga, line 101: this branch isn't 100% covered in unit test

packages/cli/src/runner/file-command-runner.ts
AhmedAbdoOrtiga: spiner

devaradhanm: try the latest version. I had fixed this. spinner was printing in new line before due to the \n.

packages/cli/src/report/report-disk-writer.spec.ts
devaradhanm: i did not make any changes to your test. just made changes to return the file name instead of file path & validate the file name is being returned. In this test, you dont have any special character anyway.

packages/cli/src/runner/file-command-runner.ts
devaradhanm: check the gif file in the PR

packages/cli/src/runner/file-command-runner.ts
devaradhanm: added tests. i had to refactor the tests to make sure that we validate the expected summary data. (this was not validated before even though coverage was 100%)

packages/cli/src/report/report-disk-writer.spec.ts
AhmedAbdoOrtiga: I know this is outta scope and it was missed in another review, but it's a tiny change and I commented about since you touched it here, anyway it could be done in another PR

PR #795, opened 2020-05-08
fix: Check response code before checking page content type

AhmedAbdoOrtiga:
#### Description of changes

- For a case like internal server error, the response code will be 500 and the page content will not be html.
- If we check the content first as we used to do, the error will be 'InvalidContentType' which is wrong.
- It should be 'HttpErrorCode', so if we check the reponse code first, we should be able to get the right error code.

#### Pull request checklist

### DIFF
--- a/packages/cli/src/scanner/page.ts
+++ b/packages/cli/src/scanner/page.ts
@@ -56,7 +56,7 @@ export class Page {
         }
 
         if (!response.ok()) {
-            console.log('url navigation returned failed response', { statusCode: response.status().toString() });
+            console.log('The URL navigation returned an unsuccessful response code', { statusCode: response.status().toString() });
 
             return {
                 error: {
--- a/packages/cli/src/scanner/page.ts
+++ b/packages/cli/src/scanner/page.ts
@@ -55,6 +55,18 @@ export class Page {
             return { error: this.getScanErrorFromNavigationFailure((err as Error).message) };
         }
 
+        if (!response.ok()) {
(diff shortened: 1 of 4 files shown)


packages/scanner/src/page.ts
```diff
@@ -53,28 +53,28 @@ export class Page {
             return { error: this.getScanErrorFromNavigationFailure((err as Error).message), pageResponseCode: undefined };
         }
 
-        if (!this.isHtmlPage(response)) {
-            const contentType = this.getContentType(response.headers());
-
-            this.log(LogLevel.error, url, 'The URL returned non-HTML content', { contentType: contentType });
+        if (!response.ok()) {
+            this.log(LogLevel.error, url, 'url navigation returned failed response', { statusCode: response.status().toString() });
```
lamaks, line 57: NIT: `The URL navigation returned an unsuccessful response code`

packages/cli/src/scanner/page.ts
```diff
@@ -55,29 +55,29 @@ export class Page {
             return { error: this.getScanErrorFromNavigationFailure((err as Error).message) };
         }
 
-        if (!this.isHtmlPage(response)) {
-            const contentType = this.getContentType(response.headers());
-
-            console.log('The URL returned non-HTML content', { contentType: contentType });
+        if (!response.ok()) {
+            console.log('url navigation returned failed response', { statusCode: response.status().toString() });
```
lamaks, line 59: NIT: `The URL navigation returned an unsuccessful response code`

PR #858, opened 2020-06-03
Added batch task retry constrain. Reduce task data retention time.

### LINKED ISSUE #859: [Bug] Enable scan batch task retry option
Enable job manager created scan task retry option to mitigate transient issues

lamaks:
#### Description of changes

Added job manager created task retry count and reduce data retention time

#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #859,#860


packages/common/src/configuration/service-configuration.ts
```diff
@@ -141,6 +143,17 @@ export class ServiceConfiguration {
                     default: 5,
                     doc: 'Timeout value after which the task has to be terminated',
                 },
+                retentionTimeInDays: {
+                    format: 'int',
+                    default: 3,
+                    doc:
+                        'The minimum time to retain the working directory for the task on the compute node where it ran, from the time it completes execution. After this time, the Batch service may delete the working directory and all its contents.',
+                },
+                maxTaskRetryCount: {
```
manishsat, line 152: what is the default value when we don't specify any value?

manishsat: No retry if nothing is specified and it keep working directory for 7 days if nothing is soecified.
https://docs.microsoft.com/en-us/azure/batch/batch-task-fail-event
https://docs.microsoft.com/en-us/rest/api/batchservice/task/add#taskconstraints

packages/azure-services/src/azure-batch/batch-task-config-generator.spec.ts
```diff
@@ -36,7 +36,12 @@ describe(BatchTaskConfigGenerator, () => {
         const commandLine = 'command line1';
         const environmentSettings = 'environment settings1' as any;
         const resourceFiles = 'resource files1' as any;
-        const defaultTaskConfig: TaskRuntimeConfig = { taskTimeoutInMinutes: 2, exitOnComplete: false };
+        const defaultTaskConfig: TaskRuntimeConfig = {
+            taskTimeoutInMinutes: 2,
+            exitOnComplete: false,
```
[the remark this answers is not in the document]
lamaks: It is used in a code from other place like https://github.com/microsoft/accessibility-insights-service/blob/9fbc5c5c675946cf41da9c578f85f8415ba0dc51/packages/service-library/src/process-entry-point-base.ts#L59

packages/common/src/configuration/service-configuration.ts
lamaks: Yes default is no retry and 7 days to keep working files

packages/azure-services/src/azure-batch/batch-task-config-generator.ts
```diff
@@ -12,14 +11,18 @@ export class BatchTaskConfigGenerator {
...
-        const maxTaskDurationInMinutes = (await this.batchTaskCommandLineProvider.getDefaultTaskConfig()).taskTimeoutInMinutes;
+        const batchTaskConfig = await this.batchTaskCommandLineProvider.getDefaultTaskConfig();
 
         return {
             id: taskId,
             commandLine: commandLine,
             resourceFiles: this.batchTaskCommandLineProvider.getResourceFiles(),
             environmentSettings: this.batchTaskCommandLineProvider.getAllEnvironmentSettings(messageText),
-            constraints: { maxWallClockTime: moment.duration({ minute: maxTaskDurationInMinutes }).toISOString() },
+            constraints: {
```
AhmedAbdoOrtiga, line 21: I was wondering why are we adding this? because if some task failed because of timeout, should we increase the timeout too?

lamaks: We are decreasing time to keep task work files on a VM host from 7 to 3 days and adding retry to handle case when task was not able to finish normally ie task has crashed abnormally.

PR #922, opened 2020-07-09
Implement error serializations

lamaks:
#### Description of changes

Implement error serializations to include non-enumerable object properties. The common JSON.stringify() construct will return an empty string like `{}`. Use dedicated npm package to serialize an error to include all properties.

#### Pull request checklist


lamaks:
> 1- Can you fix the Codecov (UTs)
> 2- When you say "Use dedicated npm package to serialize an error to include all properties." in the description, I'm curios what properties, can you give an example?

There is no regression in % of code coverage. The codecove complains when files w/o UTs originally was modified.

The issue with Error JSON serialization there are no any properties being serialized at all. The log will have only an empty object {} printed as string. The npm package will output similar instead: 

```
{ name: 'SyntaxError',
  message: 'Unexpected token T in JSON at position 0',
  stack:
   'SyntaxError: Unexpected token T in JSON at position 0\n '}
```

PR #943, opened 2020-07-27
chore: restrict azure function access to APIM IP address

### LINKED ISSUE #495: [Bug] Restrict access to Azure Web API instance to API Management IP range only
Restrict access to Azure Web API instance to Azure API Management service 's IP range only

https://docs.microsoft.com/en-us/azure/app-service/app-service-ip-restrictions

pownkel:
#### Description of changes

After this change, attempting to access the function app directly results in a 403 access denied

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #495

### DIFF
--- a/packages/resource-deployment/scripts/function-app-create.sh
+++ b/packages/resource-deployment/scripts/function-app-create.sh
@@ -144,7 +144,8 @@ deployFunctionApp() {
 }
 
 function deployWebApiFunction {
-    deployFunctionApp "web-api-allyfuncapp" "${0%/*}/../templates/function-web-api-app-template.json" "$webApiFuncAppName" "clientId=$webApiAdClientId"
+    apimIpAddress=$(az apim show --name "$apiManagementName" --resource-group "$resourceGroupName" --query "publicIpAddresses" -o tsv)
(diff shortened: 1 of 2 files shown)


packages/resource-deployment/scripts/function-app-create.sh
```diff
@@ -151,6 +151,18 @@ function deployWebWorkersFunction {
     deployFunctionApp "web-workers-allyfuncapp" "${0%/*}/../templates/function-web-workers-app-template.json" "$webWorkersFuncAppName"
 }
 
+function restrictWebApiAccess {
```
JGibson2019, line 154: Just for learning purposes: is there a particular reason y'all opted to edit programmatically rather than leverage the widget in Azure portal? Or does the Azure portal solution not support everything we need to fix the issue?

pownkel: We deploy new resource groups a lot for testing purposes, and it's easier if everything we need gets set up by the deployment scripts without having to go to the portal and change settings manually (plus it helps keep everything documented)

PR #962, opened 2020-08-15
Add debug .env file build script. Update request queue query to support multi partitions

lamaks:
#### Description of changes

Add debug .env file build script. Update request queue query to support multi partitions when partition returns empty result.

#### Pull request checklist


README.md
```diff
@@ -94,16 +94,15 @@ This project can be run in an Azure environment and can be set up easily using t
 
 To debug packages locally follow the generic steps below.
 
-1.  Complete deployment of Azure resources on your test subscription.
-2.  Create `.env` plain text file under package root folder with environment variables set required for the package to run. For instance, refer to the batch task schedule `url-scan-schedule.template.json` [template](https://github.com/microsoft/accessibility-insights-service/tree/master/packages/resource-deployment/templates) configuration for a list of common environment settings. There is a pre-built [template](https://github.com/microsoft/accessibility-insights-service/blob/master/packages/resource-deployment/.env.template) file that can be used as well. The `.env` file format:
+1.  Complete deployment of Azure resources in your test subscription.
+2.  Run the following script to create the `.env` plain text file under package root folder with environment variables required for the package to run locally:
```
AhmedAbdoOrtiga, line 98: could this be a part of the install script?

AhmedAbdoOrtiga: if the environment is "dev"

lamaks: not really, the script will grant an access for service deployment to local user who runs the script under that we do not do when install. The .env file is still specific per package like runner and sender batch jobs. To support that we need to generate one .env per package. We can do multiple file generation but not as part of install script process.

PR #969, opened 2020-08-17
feat(port-crawler-to-cli): create crawler library

pownkel:
#### Description of changes

- Move enough of the experimental crawler library (from branch users/mlaikine/crawl-api-2) to master to create a simple apify crawler to scan pages with axe
- Add unit tests for newly added classes

#### Pull request checklist


packages/crawler/src/storage/local-data-store.ts
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License.
+import Apify from 'apify';
+
+import { DataStore } from './store-types';
+
+export class LocalDataStore implements DataStore {
+    constructor(public readonly storeName: string, private datasetStore?: Apify.Dataset, private readonly apifyObj: typeof Apify = Apify) {}
+
+    public async pushData(data: object | object[]): Promise<void> {
+        await this.open().then(async () => {
```
lamaks, line 11: The 
```
await this.open();
await this.datasetStore.pushData(data);
```
should be sufficient without then() chain

packages/crawler/src/storage/local-blob-store.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export class LocalBlobStore implements BlobStore {
+    constructor(
+        public readonly storeName: string,
+        private keyValueStore?: Apify.KeyValueStore,
+        private readonly apifyObj: typeof Apify = Apify,
+    ) {}
+
+    public async setValue(key: string, value: string | Object, options?: { contentType?: string }): Promise<void> {
+        await this.open().then(async () => {
```
lamaks, line 14: why do we need to use then() and await on promise chain?

packages/crawler/src/page-scanner.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Page } from 'puppeteer';
+
+export interface ScanResult {
+    axeResults: AxeResults;
+    report?: Report;
+}
+
+export type AxePuppeteerFactory = (page: Page) => AxePuppeteer;
+
+const createAxePuppeteer: AxePuppeteerFactory = (page: Page) => new AxePuppeteer(page);
```
lamaks, line 15: can move the `(page: Page) => new AxePuppeteer(page)` to ctor directly

packages/crawler/src/page-scanner.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export class PageScanner {
+    // reporterFactory should be instantiated only once per app life cycle.
+    // Creating reporterFactory instance multiple times will result Office Fabric
+    // warning message: `Applications should only call registerIcons for any given icon once.`
+
+    public constructor(
+        private readonly reporter: Reporter = reporterFactory(),
```
lamaks, line 23: the above comment asking to keep `reporterFactory()` as singleton

packages/crawler/src/crawler-engine.ts
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import Apify from 'apify';
+// @ts-ignore
+import * as cheerio from 'cheerio';
+import { Logger } from 'logger';
```
lamaks, line 6: I would delay adding loger; otherwise convert all `console.log()` to use logger.

packages/crawler/src/page-operations/accessibility-scan-operation.ts
```diff
@@ -0,0 +1,20 @@
...
+    constructor(private readonly scanner = new PageScanner()) {}
+
+    public async run(page: Page, id: string, keyValueStore: BlobStore): Promise<void> {
+        const scanResult = await this.scanner.scan(page);
+
+        await keyValueStore.setValue(`${id}.axe`, scanResult.axeResults);
+        await keyValueStore.setValue(`${id}.report`, scanResult.report.asHTML(), { contentType: 'text/html' });
+
+        if (scanResult.axeResults.violations.length > 0) {
+            console.log(`Found ${scanResult.axeResults.violations.length} accessibility issues on page ${page.url()}`);
```
lamaks, line 17: Logger or console?

packages/crawler/src/page-processors/page-processor-base.ts
```diff
@@ -0,0 +1,76 @@
...
+import { ScanData } from '../scan-data';
+import { LocalBlobStore } from '../storage/local-blob-store';
+import { LocalDataStore } from '../storage/local-data-store';
+import { BlobStore, DataStore, scanResultStorageName } from '../storage/store-types';
+import { PageProcessorHelper } from './page-processor-helper';
+
+export interface PageProcessorOptions {
+    baseUrl: string;
+    requestQueue: Apify.RequestQueue;
+    logger: Logger;
```
lamaks, line 15: `Logger` is not logically a part of page processor options. We do not want to have `PageProcessorOptions` as bucket for everything

packages/crawler/src/page-processors/page-processor-helper.ts
```diff
@@ -0,0 +1,50 @@
...
+import { Page } from 'puppeteer';
+import { ScanData } from '../scan-data';
+import { BlobStore } from '../storage/store-types';
+
+export type PartialScanData = {
+    url: string;
+    id: string;
+} & Partial<ScanData>;
+
+export class PageProcessorHelper {
```
lamaks, line 14: why we need helper class when two functions below are used only from processors and were part of abstract base class?

packages/crawler/src/page-processors/page-processor-base.ts
```diff
@@ -0,0 +1,76 @@
...
+    public abstract pageProcessor: Apify.PuppeteerHandlePage;
+
+    /**
+     * Timeout in which page navigation needs to finish, in seconds.
+     */
+    public gotoTimeoutSecs = 30;
+
+    public constructor(
+        protected readonly requestQueue: Apify.RequestQueue,
+        protected readonly logger: Logger,
```
lamaks, line 42: NIT most significant parameter comes first i.e. logger, "helper" comes last

packages/crawler/src/page-processors/page-processor-factory.ts
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { PageProcessor, PageProcessorOptions } from './page-processor-base';
+
+export interface PageProcessorFactory {
```
lamaks, line 5: why we are moving away from objects to functional programming? we can have interface and abstract class at the same time. not clear why need to create get-discovery-pattern.ts with functions and drop abstract class.

packages/crawler/src/page-processors/page-processor-helper.ts
pownkel: Since both page processors (classic and simulator) use these methods, I thought it would be easier to move them into a separate object that we can easily mock for testing, instead of testing for the behavior of base class methods (which would mean changing multiple tests if the base class changes)

packages/crawler/src/page-processors/page-processor-factory.ts
pownkel: This was another thing I moved for ease of testing, but since without this function the PageProcessorFactories are so simple that they don't necessarily need their own tests, I guess it isn't needed.

packages/crawler/src/storage/local-data-store.ts
AhmedAbdoOrtiga: then does not stops the execution flow and runs next function in the chain only when the previous promise is finished, maybe it's not needed as you mentioned but I usually prefer it, we can revert it tho

packages/crawler/src/storage/local-blob-store.ts
AhmedAbdoOrtiga: same like above

PR #971, opened 2020-08-19
feat(port-crawler-to-cli): Add simulator page processor

AhmedAbdoOrtiga:
#### Description of changes
This PR adds another page processor that simulates user click on elements that match to the specified selectors, if the simulate arg is true, the simuator page will run using the passed selectors in the args, the default is button.

#### Pull request checklist


packages/crawler/src/crawler/crawler-engine.spec.ts
```diff
@@ -82,14 +96,15 @@ describe(CrawlerEngine, () => {
     });
 
     afterEach(() => {
-        pageProcessorFactoryMock.verifyAll();
+        // pageProcessorFactoryMock.verifyAll();
+        // crawlerFactoryMock.verifyAll();
+        // puppeteerCrawlerMock.verifyAll();
```
pownkel, line 101: Why are these commented out?

.vscode/launch.json
```diff
@@ -5,15 +5,20 @@
     "version": "0.2.0",
     "configurations": [
         {
+            "name": "Debug current test file in VSCode",
```
lamaks, line 8: This is global to solution change. Does this work across all packages?

packages/crawler/src/apify-resources/resource-creator.ts
```diff
@@ -2,13 +2,12 @@
 // Licensed under the MIT License.
 import Apify from 'apify';
 
-export interface CrawlerFactory {
+export interface ResourceCreator {
```
lamaks, line 5: Too generic name, rename to ApifyResourceCreator?

packages/crawler/src/index.ts
```diff
@@ -4,4 +4,58 @@
...
+    const crawlerEngine: CrawlerEngine = new CrawlerEngine();
+
+    await crawlerEngine.start({
+        baseUrl: args.url,
+        simulate: args.simulate,
+        selectors: args.selectors,
+        localOutputDir: args.output,
+    });
+})().catch((error) => {
+    console.log('Exception: ', error);
```
lamaks, line 59: serialize error with `System.serializeError`

.vscode/launch.json
AhmedAbdoOrtiga: It works when you debug unit test in vscode by clicking f5 for any unit test file.

packages/crawler/src/apify-resources/resource-creator.ts
AhmedAbdoOrtiga: This is the interface, that's why it's generic, the implementation of it in the same file is ApifyResourceCreator which uses apify resources, later we can use something like MemoryResourceCreator for using the memory as a resource

packages/crawler/src/crawler/crawler-engine.spec.ts
AhmedAbdoOrtiga: fixed, was working on it

packages/crawler/src/index.ts
AhmedAbdoOrtiga: fixed

PR #972, opened 2020-08-20
feat(port-crawler-to-cli): Error logging

pownkel:
#### Description of changes

- Use Logger instead of console.log
- Log the last error for a request in a file named [requestId].err.txt in the same directory as the scan results
- fix issues with env variables not set correctly

When a page fails to open or scan, the request is put in the back of the queue for retry. The .err.txt file for that request will contain the last error from scanning that page, even if it succeeded on retry.

Sample .err.txt file contents:
`
Error at URL https://www.google.com/finance: Evaluation failed: TypeError: Cannot read property 'configure' of undefined
    at runAxe (__puppeteer_evaluation_script__:10:16)
`

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1760046


packages/crawler/src/crawler/crawler-engine.ts
```diff
@@ -16,38 +16,40 @@ export type ApifyMainFunc = (userFunc: Apify.UserFunc) => void;
...
-        const pageProcessor = this.pageProcessorFactory.createPageProcessor({
-            requestQueue,
-            crawlerRunOptions,
-        });
+        const pageProcessor = this.pageProcessorFactory.createPageProcessor(
+            {
+                requestQueue,
+                crawlerRunOptions,
+            },
+            this.logger,
```
AhmedAbdoOrtiga, line 37: can you make this a Singleton class and pass it in the parameter of each class? I'm planning to switch to inverse dependence injection later if we got time at the end of the feature.

packages/crawler/src/crawler/crawler-engine.ts
```diff
@@ -16,38 +16,40 @@ export type ApifyMainFunc = (userFunc: Apify.UserFunc) => void;
...
                 gotoFunction: pageProcessor.gotoFunction,
                 handleFailedRequestFunction: pageProcessor.pageErrorProcessor,
             });
             await crawler.run();
         });
     }
 
-    private setLocalOutputDir(outputDir: string): void {
-        setApifyEnvVars({ APIFY_LOCAL_STORAGE_DIR: outputDir });
+    private setEnvVars(localOutputDir?: string): void {
```
AhmedAbdoOrtiga, line 52: can you move this to crawlerConfiguration and rename it to "setOutputDirectory"?
if you think it's out of scope, we can do it in another PR

packages/crawler/src/index.ts
```diff
@@ -48,7 +51,11 @@ interface Arguments {
         })
         .describe('help', 'Print command line options').argv as unknown) as Arguments;
 
-    const crawlerEngine: CrawlerEngine = new CrawlerEngine();
+    const serviceConfig = new ServiceConfiguration();
```
AhmedAbdoOrtiga, line 54: same, this might need to be a singltone logger

packages/crawler/src/page-processors/page-processor-base.ts
```diff
@@ -14,43 +15,62 @@ export type PartialScanData = {
...
      */
     public gotoFunction: Apify.PuppeteerGoto = async (inputs: Apify.PuppeteerGotoInputs) => {
-        return this.gotoExtended(inputs.page, inputs.request, {
-            waitUntil: 'networkidle0',
-            timeout: this.gotoTimeoutSecs * 1000,
-        });
+        try {
+            return await this.gotoExtended(inputs.page, inputs.request, {
+                waitUntil: 'networkidle0',
+                timeout: this.gotoTimeoutSecs * 1000,
```
AhmedAbdoOrtiga, line 66: do you think it'd be better if we set time out in the "options: Apify.PuppeteerCrawlerOptions" in "this.crawlerFactory.createPuppeteerCrawler" in crawler-engine.ts?

PR #977, opened 2020-08-21
feat(port-crawler-to-cli): Use InversifyJS for dependency injection

AhmedAbdoOrtiga:
#### Description of changes
Use InversifyJS for dependency injection to follow same pattern for all other packages

#### Pull request checklist


packages/crawler/src/crawler-entry-point.ts
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { Container } from 'inversify';
+import { GlobalLogger } from 'logger';
+import { CrawlerRunOptions } from '.';
```
lamaks, line 6: add file name to import

packages/crawler/src/page-processors/simulator-page-processor.ts
```diff
@@ -1,39 +1,42 @@
...
 // tslint:disable: no-unsafe-any
 export class SimulatorPageProcessor extends PageProcessorBase {
     public constructor(
-        protected readonly logger: Logger,
+        @inject(EnqueueActiveElementsOperation) protected readonly enqueueActiveElementsOp: EnqueueActiveElementsOperation,
+        @inject(ClickElementOperation) protected readonly clickElementOp: ClickElementOperation,
+        @inject(AccessibilityScanOperation) protected readonly accessibilityScanOp: AccessibilityScanOperation,
+        @inject(LocalDataStore) protected readonly dataStore: DataStore,
+        @inject(LocalBlobStore) protected readonly blobStore: BlobStore,
+        @inject(GlobalLogger) private readonly loggerSimulator: GlobalLogger,
```
lamaks, line 26: may just `logger` instead to avoid confusion about special logger functionality

packages/crawler/src/scanner-operations/page-scanner.ts
```diff
@@ -1,31 +1,27 @@
...
     public constructor(
-        private readonly reporter: Reporter = singeltonReporter,
-        private readonly createAxePuppeteerFunc: AxePuppeteerFactory = (page: Page) => new AxePuppeteer(page),
+        @inject('ReporterFactory') private readonly reporter: Reporter,
+        @inject(AxePuppeteerFactory) private readonly axePuppeteerFactory: AxePuppeteerFactory,
     ) {}
 
     public async scan(page: Page): Promise<ScanResult> {
-        const axePuppeteer: AxePuppeteer = this.createAxePuppeteerFunc(page);
+        // tslint:disable-next-line: no-unsafe-any
```
lamaks, line 23: seems unnecessary lint comment, can you see why we need this for typed vars

packages/crawler/src/page-processors/page-processor-factory.ts
```diff
@@ -1,19 +1,36 @@
...
+import { LocalBlobStore } from '../storage/local-blob-store';
+import { LocalDataStore } from '../storage/local-data-store';
 import { PageProcessorOptions } from '../types/run-options';
+import { ActiveElementsFinder } from '../utility/active-elements-finder';
 import { ClassicPageProcessor } from './classic-page-processor';
 import { PageProcessor } from './page-processor-base';
 import { SimulatorPageProcessor } from './simulator-page-processor';
 
+@injectable()
 export class PageProcessorFactory {
```
lamaks, line 21: This is good candidate to move to container as well

packages/crawler/src/page-processors/simulator-page-processor.ts
AhmedAbdoOrtiga: logger isn't injectable

AhmedAbdoOrtiga: also I'll add a task to create a console logger so it doesn't need to take service configuration and print this unnecessary information

packages/crawler/src/page-processors/page-processor-factory.ts
AhmedAbdoOrtiga: it would've been good if there was no input parameter, I know still we can do it but it'll be more complicated

packages/crawler/src/setup-crawler-container.ts
```diff
@@ -0,0 +1,48 @@
...
+    container.bind(ClassicPageProcessor).toSelf().inSingletonScope();
+    container.bind(PageProcessorFactory).toSelf().inSingletonScope();
+    container.bind(SimulatorPageProcessor).toSelf().inSingletonScope();
+    container.bind(PageProcessorBase).toSelf().inSingletonScope();
+    container.bind(PageScanner).toSelf().inSingletonScope();
+    container.bind('ReporterFactory').toConstantValue(reporterFactory);
+    container.bind<AxePuppeteerFactory>(AxePuppeteerFactory).toSelf().inSingletonScope();
+
+    container.bind(LocalBlobStore).toSelf().inSingletonScope();
+    container.bind(LocalDataStore).toSelf().inSingletonScope();
```
lamaks, line 43: Do not really need to have every class bind as singleton unless we have to. @injectable on a class will suffice.

AhmedAbdoOrtiga: it makes sense for most of the classes to be bind to singletonScope, do you think any of them shouldn't?

packages/crawler/src/page-processors/simulator-page-processor.ts
lamaks: I mean just to rename to logger instead

packages/crawler/src/setup-crawler-container.ts
lamaks: I just think `CrawlerConfiguration` can only be one, rest is concerning to me as we need to ensure that this will not break logic like for example have Page as singleton as result of making all classes as singleton

packages/crawler/src/page-processors/simulator-page-processor.ts
AhmedAbdoOrtiga: if I use the same name, I'll get error 
"Class 'SimulatorPageProcessor' incorrectly extends base class 'PageProcessorBase'.
  Property 'logger' is private in type 'SimulatorPageProcessor' but not in type 'PageProcessorBase'."

same for enqueueLinksSimulator and gotoExtendedSimulator

lamaks: to resolve this just make same var scope in both base and derived classes like private or protected

lamaks (dismissed):

PR #989, opened 2020-08-25
feat(port-crawler-to-cli): get roorUrl and validate baseUrl

AhmedAbdoOrtiga:
#### Description of changes

This PR addresses:
  - Make sure that the base URL doesn't have query parameters.
  - Get the rootUrl where the  position of the last '/' in the path defines the sibling crawl scope

    - Example1
       Root URL: http://foo.com/a/b1 
       Will crawl: http://foo.com/a/b1/c & http://foo.com/a/b2
       Will not crawl: http://foo.com/a2

    - Example2
       Root URL: http://foo.com/a/b1/ 
       Will crawl: http://foo.com/a/b1/c
       Will not crawl: http://foo.com/a2 & https://foo.com/a/b2

#### Pull request checklist


packages/crawler/src/utility/url-processor.ts
```diff
@@ -0,0 +1,19 @@
+import { injectable } from 'inversify';
+
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
lamaks, line 3: NIT move copyright to the top of the file

packages/crawler/src/utility/url-processor.ts
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT License.
+
+@injectable()
+export class URLProcessor {
```
lamaks, line 7: Use node url lib instead
See https://github.com/microsoft/accessibility-insights-service/blob/master/packages/common/src/system/url.ts

PR #993, opened 2020-08-25
Refactor web driver

lamaks:
#### Description of changes

Refactor web driver and page scanner to be in sync with service implementation and address false positive color-contrast issue.

#### Pull request checklist

### DIFF
--- a/packages/cli/src/scanner/page.ts
+++ b/packages/cli/src/scanner/page.ts
@@ -16,6 +16,9 @@ export class Page {
     public puppeteerPage: Puppeteer.Page;
     public browser: Puppeteer.Browser;
 
+    private readonly pageNavigationTimeoutMsecs = 15000;
+    private readonly pageRenderingTimeoutMsecs = 5000;
+
     constructor(
         @inject(AxePuppeteerFactory) private readonly axePuppeteerFactory: AxePuppeteerFactory,
         @inject(WebDriver) private readonly webDriver: WebDriver,
@@ -26,6 +29,7 @@ export class Page {
             this.browser = await this.webDriver.launch(chromePath);
         }
         this.puppeteerPage = await this.browser.newPage();
+        await this.puppeteerPage.setUserAgent(this.webDriver.userAgent);
     }
 
     public async enableBypassCSP(): Promise<void> {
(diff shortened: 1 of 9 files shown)


packages/cli/src/web-driver/web-driver.ts
```diff
@@ -1,27 +1,23 @@
...
 
     constructor(private readonly puppeteer: typeof Puppeteer = Puppeteer) {}
 
     public async launch(chromePath?: string): Promise<Puppeteer.Browser> {
-        let launchOption: Puppeteer.LaunchOptions;
-        launchOption = {
+        this.browser = await this.puppeteer.launch({
             headless: true,
-            timeout: 30000,
+            executablePath: chromePath,
```
AhmedAbdoOrtiga, line 16: if chromePath is null, will puppeteer pick up the default or set it to null?

packages/cli/src/web-driver/web-driver.spec.ts
```diff
@@ -60,4 +63,11 @@ describe('WebDriver', () => {
         expect(browser).toEqual(puppeteerBrowserMock);
         puppeteerLaunchMock.verifyAll();
     });
+
+    it('should update user agent string', async () => {
+        await testSubject.launch();
+
+        expect(testSubject.userAgent).toEqual('Chrome user agent string');
+        puppeteerLaunchMock.verifyAll();
```
AhmedAbdoOrtiga, line 71: can we move this to afterEach

packages/cli/src/scanner/page.ts
```diff
@@ -100,17 +112,45 @@ export class Page {
...
         }
 
         return scanResults;
     }
 
-    public async close(): Promise<void> {
-        if (this.puppeteerPage !== undefined) {
-            await this.puppeteerPage.close();
-            await this.webDriver.close();
+    private async waitForPageToCompleteRendering(page: Puppeteer.Page, timeoutMsecs: number): Promise<void> {
```
AhmedAbdoOrtiga, line 123: Will we have this issue in the crawler?

packages/cli/src/scanner/page.ts
```diff
@@ -39,25 +43,29 @@ export class Page {
...
-        const gotoUrlPromise = this.puppeteerPage.goto(url, { waitUntil: ['load'], timeout: 120000 });
-        const networkLoadTimeoutInMilleSec = 30000;
-        const waitForNetworkLoadPromise = this.puppeteerPage.waitForNavigation({
-            waitUntil: ['networkidle0'],
-            timeout: networkLoadTimeoutInMilleSec,
-        });
-
-        let response;
+        // separate page load and networkidle0 events to bypass network activity error
+        const gotoUrlPromise = this.puppeteerPage.goto(url, { waitUntil: 'load', timeout: this.pageNavigationTimeoutMsecs });
```
AhmedAbdoOrtiga, line 47: so if we wait in here, why do we still need 'waitForPageToCompleteRendering'?

packages/cli/src/web-driver/web-driver.ts
lamaks: It works as it is null by default, we do not pass any value when run app.

packages/cli/src/scanner/page.ts
lamaks: yes, this should be added to crawler as well

packages/cli/src/scanner/page.ts
lamaks: some pages will have active javascript after load event that affect page rendering and as result color contrast detection. This results in false positive accessibility issues from axe.

PR #996, opened 2020-08-26
Force axe package version when deploying in Docker container

lamaks:
#### Description of changes

Force axe package version when deploying in Docker container

#### Pull request checklist


yarn.lock
```diff
@@ -9189,6 +9194,14 @@ npm-bundled@^1.0.1:
   dependencies:
     npm-normalize-package-bin "^1.0.1"
 
+npm-force-resolutions@^0.0.3:
```
DaveTryon, line 9197: The package page (https://www.npmjs.com/package/npm-force-resolutions) describes npm-force-resolutions as an option of last resort. Would it make sense to consider modifying the packages that we exclude, so that the resolutions in the top-level packages.json come into play? I don't have a strong preference either way, but wanted to float the alternative approach.

lamaks: The top level config will not work in case of container preparation since we use npm instead of yarn

PR #1028, opened 2020-09-17
feat(cli-summary-report): Create summary reports (html summary and error log) for batch scanning 

AhmedAbdoOrtiga:
#### Description of changes

This PR add the following feature for file scanning:
- Create html summary report.
- Create error log.
- Update CLI notification.

#### Pull request checklist

### DIFF
--- a/packages/cli/src/runner/crawler-command-runner.ts
+++ b/packages/cli/src/runner/crawler-command-runner.ts
@@ -12,7 +12,6 @@ import { CommandRunner } from './command-runner';
 
 @injectable()
 export class CrawlerCommandRunner implements CommandRunner {
-    private scanResult: ScanResults;
 
     constructor(
         @inject(CrawlerEntryPoint) private readonly crawlerEntryPoint: CrawlerEntryPoint,
@@ -27,7 +26,7 @@ export class CrawlerCommandRunner implements CommandRunner {
 
         console.log(`Crawling and scanning page ${scanArguments.url}`);
 
-        this.scanResult = await this.crawlerEntryPoint.crawl({
+        const scanResult = await this.crawlerEntryPoint.crawl({
             baseUrl: scanArguments.url,
             simulate: scanArguments.simulate,
             selectors: scanArguments.selectors,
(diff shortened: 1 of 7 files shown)


packages/cli/src/runner/crawler-command-runner.ts
```diff
@@ -25,7 +27,7 @@ export class CrawlerCommandRunner implements CommandRunner {
 
         console.log(`Crawling and scanning page ${scanArguments.url}`);
 
-        const scanResult = await this.crawlerEntryPoint.crawl({
+        this.scanResult = await this.crawlerEntryPoint.crawl({
```
waabid, line 30: Why not just pass this as a parameter to the generate summary reports method? Would likely help with a future re-factor as well, if we decide to break apart this class.

packages/cli/src/runner/crawler-command-runner.ts
```diff
@@ -42,35 +44,46 @@ export class CrawlerCommandRunner implements CommandRunner {
...
+        startDateNumber: number,
+        endDate: Date,
+        endDateNumber: number,
+    ): Promise<void> {
         const durationSeconds = (endDateNumber - startDateNumber) / 1000;
         console.log(`Done in ${durationSeconds}s`);
 
-        const scannedPagesCount = scanResult.summaryScanResults.failed.length + scanResult.summaryScanResults.passed.length;
-        const discoveredPagesCount = scannedPagesCount + scanResult.summaryScanResults.unscannable.length;
+        const scannedPagesCount = this.scanResult.summaryScanResults.failed.length + this.scanResult.summaryScanResults.passed.length;
```
waabid, line 61: Should we be null checking the scan result? As in, are we always sure it will have these properties and what not?

AhmedAbdoOrtiga: we don't need to check at least for now, if the crawler package has no results, it will return an object empty with arrays of length 0 eg:
scanResult : {
summaryScanResults: {
failed:[],
passed:[],
unscannable:[],
}
}

lamaks (approved):

PR #1036, opened 2020-09-18
Add continue option to crawler cli tool

lamaks:
#### Description of changes

Add continue option to crawler cli tool
Refactor store of a scan metadata

#### Pull request checklist

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -30,8 +30,6 @@ function readScanArguments(): ScanArguments {
 
     return (yargs
         .wrap(yargs.terminalWidth())
-        .implies('crawl', 'url')
-        .implies('selectors', 'simulate')
         .options({
             crawl: {
                 type: 'boolean',
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -2,7 +2,6 @@
 
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-
 import 'reflect-metadata';
 // tslint:disable-next-line: no-import-side-effect
 import './module-name-mapper';
@@ -27,8 +26,12 @@ import { setupCliContainer } from './setup-cli-container';
 });
 
 function readScanArguments(): ScanArguments {
+    const defaultOutputDir = 'ai_scan_cli_output';
+
     return (yargs
         .wrap(yargs.terminalWidth())
+        .implies('crawl', 'url')
+        .implies('selectors', 'simulate')
         .options({
             crawl: {
                 type: 'boolean',
@@ -51,8 +54,8 @@ function readScanArguments(): ScanArguments {
             },
             output: {
                 type: 'string',
-                describe: `Output directory. Defaults to the value of APIFY_LOCAL_STORAGE_DIR, if set, or ./crawler_storage, if not.`,
-                default: 'ai_scan_cli_output',
+                describe: `Output directory. Defaults to the value of APIFY_LOCAL_STORAGE_DIR, if set, or ./${defaultOutputDir}, if not.`,
+                default: defaultOutputDir,
             },
             maxUrls: {
                 type: 'number',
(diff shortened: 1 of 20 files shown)


packages/cli/src/runner/crawler-command-runner.ts
```diff
@@ -46,6 +51,18 @@ export class CrawlerCommandRunner implements CommandRunner {
         await this.generateSummaryReports(scanArguments, scanResult, startDate, startDateNumber, endDate, endDateNumber);
     }
 
+    private canRun(scanArguments: ScanArguments): boolean {
+        if (this.filesystem.existsSync(scanArguments.output) && scanArguments.restart === false && scanArguments.continue === false) {
```
waabid, line 55: Is there a test for this?

packages/cli/src/cli.ts
```diff
@@ -101,6 +107,10 @@ function readScanArguments(): ScanArguments {
                 if (isEmpty(args.url)) {
                     throw new Error('The --url option is required.');
                 }
+
+                if (args.restart === true && args.continue === true) {
```
waabid, line 111: Seems like we don't have tests for this or the cli-client-runner changes. Is this intentional/necessary?

packages/cli/src/runner/crawler-command-runner.ts
lamaks: added test

packages/cli/src/cli.ts
lamaks: will add later

packages/cli/src/report/report-disk-writer.ts
```diff
@@ -6,59 +6,68 @@ import filenamifyUrl from 'filenamify-url';
 import * as fs from 'fs';
 import { injectable } from 'inversify';
 import { isEmpty } from 'lodash';
+import normalizePath from 'normalize-path';
 import * as path from 'path';
 import { ReportFormats } from './report-formats';
 
+// tslint:disable: no-parameter-reassignment
```
waabid, line 13: Why is this required?

packages/cli/src/runner/crawler-command-runner.spec.ts
```diff
@@ -3,79 +3,148 @@
...
-    const testUrl = 'http://www.bing.com';
-    const testInput: ScanArguments = { url: testUrl, output: '/users/xyz' };
-    // tslint:disable-next-line: mocha-no-side-effect-code
+
     beforeEach(() => {
+        testInput = { url: testUrl, output: './dir' };
+        crawlerOption = {
+            baseUrl: testInput.url,
+            localOutputDir: testInput.output,
+            existingUrls: undefined,
```
waabid, line 46: why do these have to be defined? why not just cast the object instead?

lamaks: example?

packages/cli/src/report/report-disk-writer.ts
lamaks: left over

packages/cli/src/report/report-disk-writer.spec.ts
```diff
@@ -31,37 +32,67 @@ describe('ReportDiskWriter', () => {
         testSubject = new ReportDiskWriter(fsMock.object, pathMock.object);
     });
 
+    afterEach(() => {
+        fsMock.verifyAll();
```
waabid, line 36: Not sure what the pattern is here but IIRC the pattern in the web project is to avoid verify in afterEach method.

packages/cli/src/runner/crawler-command-runner.spec.ts
waabid:
```

crawlerOption = {
    baseUrl: testInput.url,
    localOutputDir: testInput.output,
} as CrawlerRunOptions;
```
should do the trick.

lamaks: does not work, need to have explicit undefined

packages/cli/src/report/report-disk-writer.spec.ts
lamaks: the pattern here is to have common mocks setup to verify so they are not missed or skip. Had several cases when this helped. It will not break the ability verify locally in test  It will also verify any setup set in a test or skip if it is not set in test.

packages/cli/src/runner/crawler-command-runner.spec.ts
waabid: Hm, I just tried it on my end, and it seemed to work I didn't even need the `as` casting. Hmmmm...

waabid (approved):
LGTM, minor comments.

PR #1168, opened 2020-12-07
feat(axe-core-4.1.1): upgrade to axe-core 4.1.1 (including report support)

dbjorge:
#### Description of changes

This updates `axe-core`, `@axe-core/puppeteer`, and `accessibility-insights-report` in lockstep, including axe-core yarn resolutions.

It adds the new axe-core best-practice rules (from the rule table linked in the feature description) to the rule-exclusion list.

It also adds `accessibility-insights-report` as an external in the `web-api-scan-runner` webpack configuration; this means that rather than bundling it, the docker container will fetch the report package during its `npm install` phase. This prevents an error where the `report` package introduced a new runtime dependency which was not in turn treated as an external in the web-api-scan-runner; now, the scan runner will no longer need to care when the report package introduces new transitive deps.

#### Pull request checklist

- [x] Addresses an existing issue: Fixes 1797801
- [n/a] Added relevant unit test for your changes. (`yarn test`)
- [n/a] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] Ran precheckin (`yarn precheckin`)

PR #1178, opened 2020-12-10
feat(axe-core-4.1.1): treat *all* external node modules as external during ai-scan bundling

dbjorge:
#### Description of changes

Before this PR, the CLI's webpack configuration used a hardcoded list of "externals" when bundling the CLI bundle and the default `nodeExternals()` output as externals when bundling the index.js entry point.

Neither of these strategies are ideal:

* `nodeExternals()` doesn't understand monorepos by default; in particular, it only looks looks for modules present under the immediate package's `node_modules` folder, *not* any lifted packages under the repo-root `node_modules`. If you naively add the root-level `node_modules` as an `additionalModuleDir`, you'll find that it also does not understand *not* to treat other packages within the monorepo that are symlinked under the root `node_modules` as external, which doesn't work for us because we don't publish the intermediate monorepo packages (we intentionally bundle them).
* The hard-coded list is fragile; we have several transitive dependencies that include native (ie, arch/platform-specific) modules, which cause hard-to-detect errors if they are accidentally missed from the list (the project will build fine and may even run fine via `node dist/ai-scan-cli.js`, but when packed and used from a consumer outside the monorepo, will produce confusing errors about "missing an arch for a node-gyp-build import'.
* The hard-coded list causes us to over-bundle; right now, the distributed package bundles most dependencies (everything not in the hard-coded list) rather than expressing it as a package.json dependency, which is not great; distributing more code than we have to increases our legal and technical liability and causes consumers to install 2 copies of any dependencies we bundle that they'd have pulled down non-bundled for other reasons.

I believe a better strategy than both of the current options would be to bundle *only* other packages defined in the service monorepo and treat as external all other node_modules external to the repo. This PR implements that strategy by using `nodeExternals` with an explicit additional module directory for the lifted root-level packages and an explicit non-external-allowlist of "all monorepo packages".

The main disadvantage of this approach is that it adds a new required invariant of the CLI's package.json; from a library consumer's perspective, none of the other monorepo packages' `package.json` files exist, even though we are bundling those other packages' code into the CLI package. For library consumers to correctly understand what dependencies they need to run that bundled code, we need to include *all* bundled monorepo packages' dependencies as subsets of the CLI package dependencies, even those which the CLI package's individual code has no direct dependency on. A new test case in `package.json.spec.ts` enforces this invariant (as well as the pre-existing invariant that other monorepo packages need to be `devDependencies`, not `dependencies`).

A simpler approach that avoids both the complexity and invariants would be for us to just publish the intermediate packages to npm, rather than bundling them into the scan package; then we could probably avoid using webpack at all in the CLI package's build. I think that would probably be my preference, but I didn't want to expand the scope of the axe-core update feature all the way to "broadly changing our publishing strategy and introducing new version pinning requirements in the service repo".

#### Testing

Beyond `yarn precheckin`, manual verification steps of the scan package from a consumer's point of view:
* [x] produces a library package from `yarn run pack`
* [x] `yarn add /path/to/cli.tgz` from a test project outside the monorepo (careful of yarn caching; do `yarn cache clean` between CLI builds if you play with this locally)
* [x] `npx ai-scan --help` from non-monorepo test project as a smoke test (this is the first step that fails in #1168)
* [x] `npx ai-scan --url https://accessibilityinsights.io` and `npx ai-scan --crawl --url https://accessibilityinsights.io` from non-monorepo test project as more involved e2e tests
* [x] importing the scan package via a `index.ts` entry point (like action does) and running it that way. The test code I used was:

```ts
import 'reflect-metadata';
import { AIScanner } from 'accessibility-insights-scan';
import { Container } from 'inversify';

const container = new Container({ autoBindInjectable: true });
const scanner = container.get(AIScanner);
scanner.scan('https://accessibilityinsights.io')
    .then((results) => console.log(JSON.stringify(results, null, 2)))
    .catch((e) => console.error('AIScanner.scan threw error', e));

```

#### Pull request checklist

- [x] Addresses an existing issue: unblocks 1797818 / #1168


lisli1 (dismissed):
LGTM

pownkel (approved):
LGTM

packages/cli/src/package.json.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    it('does not include any direct non-dev dependencies on monorepo packages', () => {
+        expect(monorepoNonDevDependencies).toEqual([]);
+    });
+
+    // We bundle the other monorepo packages (eg, "common") with what we publish to npm, but
+    // we treat as external all node_modules, including those that are only included as transitive
+    // dependencies of other monorepo packages. To avoid breaking consumers, that means that we
+    // need to repeat those transitive dependencies in our own dependency list; ie, the "cli" dependency
+    // list should be a superset of the "common" dependency list.
+    it.each(monorepoDevDependencies)("is a superset of %s's dependencies", async (monrepoPackageName) => {
```
lamaks, line 26: typo monrepoPackageName

packages/cli/src/package.json.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+    // suppose:
+    //   * /packages/monorepo-foo/package.json has dependencies: { 'monorepo-bar': '*', 'external-a': '*' }
+    //   * /packages/monorepo-bar/package.json has dependencies: { 'external-b': '*' }
+    //   * /node_modules/external-a/package.json has dependencies: { 'external-c': '*' }
+    //   * external-b and external-c have no dependencies
+    //
+    // then getEdgeNonMonorepoDependencies('monorepo-foo') returns ['external-a', 'external-b']
+    async function getEdgeNonMonorepoDependencies(monorepoPackageName: string): Promise<string[]> {
```
lamaks, line 41: what is specific about `Edge` for this implementation?

packages/cli/src/package.json.spec.ts
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import 'reflect-metadata';
+import { flatten } from 'lodash';
+import { listMonorepoPackageNames } from 'common';
+import * as packageJson from '../package.json';
+
+describe('package.json dependencies', () => {
```
lamaks, line 8: How this test will fail if CLI package has missing external dependencies? The ideal output would be diff of external dependencies that are missing from the current CLI package.json

lamaks (changes requested):
Added new comments

packages/cli/src/package.json.spec.ts
dbjorge: It only includes the "edge" of the dependency graph consisting of the first layer of non-monorepo packages; ie, it does not include all transitive non-monorepo dependencies (in the example from the comment, it doesn't include 'external-c'), but it also does include some non-direct dependencies (in the example from the comment, it does include 'external-b'). Do you have a suggestion for a more intuitive name?

packages/cli/src/package.json.spec.ts
dbjorge: Updated the test to consolidate all missing deps into one output. An example of the new failure text (from removing @medv/finder and raw-body temporarily from package.json):

```
  ● package.json dependencies › includes direct dependencies for each non-monorepo dependency of its transitive monorepo dependencies

    expect(received).toBe(expected) // Object.is equality

    - Expected  - 1
    + Received  + 4

    - Array []
    + Array [
    +   "package.json needs a dependency on \"@medv/finder\": \"^2.0.0\" (required via accessibility-insights-crawler)",
    +   "package.json needs a dependency on \"raw-body\": \"^2.4.1\" (required via accessibility-insights-crawler, axe-result-converter, common, scanner-global-library)",
    + ]

      47 |             });
      48 |
    > 49 |         expect(missingDependencyExplanations).toBe([]);
```

lamaks (approved):

PR #1221, opened 2021-01-12
feat(cloud-crawler): SimpleCrawlerEngine implementation

pownkel:
#### Description of changes

Implement SimpleCrawlerEngine
- Discover Urls on the page provided
- Return list of urls discovered
- Add crawl options for puppeteer page and discoveryPatternUrl (url to be used in discovery pattern generation)

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1807310

### DIFF
--- a/packages/crawler/src/crawler/simple-crawler-engine.spec.ts
+++ b/packages/crawler/src/crawler/simple-crawler-engine.spec.ts
@@ -10,7 +10,7 @@ import { WebDriver } from 'scanner-global-library';
 
 import { IMock, It, Mock } from 'typemoq';
 import { ApifyResourceCreator } from '../apify/apify-resource-creator';
-import { RequestProcessor } from '../page-processors/request-processor';
+import { CrawlRequestProcessor } from '../page-processors/crawl-request-processor';
 import { UrlCollectionRequestProcessor } from '../page-processors/url-collection-request-processor';
 import { CrawlerRunOptions } from '../types/crawler-run-options';
 import { ApifyRequestQueueProvider } from '../types/ioc-types';
@@ -38,7 +38,7 @@ describe(SimpleCrawlerEngine, () => {
         handleRequest: () => null,
         handleFailedRequest: () => null,
         getResults: () => crawlResults,
-    } as RequestProcessor;
+    } as CrawlRequestProcessor;
 
     beforeEach(() => {
         requestQueueStub = {} as Apify.RequestQueue;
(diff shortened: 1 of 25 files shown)


packages/crawler/src/crawler/simple-crawler-engine.spec.ts
```diff
@@ -2,28 +2,164 @@
...
+        setupEnqueueLinks();
+
+        const expectedUrls: string[] = crawlResults;
 
         const discoveredUrls = await testSubject.start(crawlerRunOptions);
 
         expect(discoveredUrls).toEqual(expectedUrls);
     });
+
+    it.skip('run e2e website crawl', async () => {
```
RobGallo, line 101: was .skip purposly left in the PR?

pownkel: Yes, this test creates all the dependencies of the crawler instead of mocking them and runs it end-to-end on a real site. It's marked as skip because we probably don't want to do this in the build/test pipeline, but I left the code in for manual local testing if we need it.

packages/crawler/src/page-processors/request-processor.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import Apify from 'apify';
+
+export interface RequestProcessor {
```
RobGallo, line 6: Perhaps give this a less generic name like CrawlRequestProcessor?

packages/crawler/src/page-processors/request-processor.ts
```diff
@@ -0,0 +1,10 @@
...
+    handleRequest(inputs: Apify.HandleRequestInputs): Promise<void>;
```
RobGallo, line 7: Perhaps give this a name like handleSuccessfulRequest to make it symetrical with handleFailedRequest?

RobGallo (approved):
LGTM! A few knits which you can take or leave.

packages/crawler/src/page-processors/request-processor.ts
pownkel: I'm not sure if handleSuccessfulRequest would be an accurate name, handleFailedRequest is called on an error in handleRequest so there's no guarantee that the request is successful. Maybe renaming handleFailedRequest to handleError would be clearer?

RobGallo: How about handleRequestError?

RobGallo (approved):

PR #1237, opened 2021-01-19
feat(cloud-crawler): expose underlying puppeteer page to Runner

pownkel:
#### Description of changes

- Separate page scanning from page navigation
- Expose the underlying puppeteer page so we can access it when crawling
- Move Page from Scanner into Runner, so it can be passed to the crawler when we add that step

#### Pull request checklist

- [x] Addresses an existing issue: #1808725


brocktaylor7 (approved):
LGTM

PR #1245, opened 2021-01-20
feat(cloud-crawler): DiscoveredUrlsProcessor

pownkel:
#### Description of changes

Create DiscoveredUrlsProcessor, which takes a list of discovered URLs, filters out known URLs, and truncates the list according to the configured limit

#### Pull request checklist

- [x] Addresses an existing issue: #1808730


packages/web-api-scan-runner/src/crawler/discovered-urls-processor.ts
```diff
@@ -0,0 +1,41 @@
...
+export class DiscoveredUrlsProcessor {
+    public constructor(@inject(ServiceConfiguration) private readonly serviceConfig: ServiceConfiguration) {}
+
+    public async getProcessedUrls(discoveredUrls: string[], websiteScanResult?: WebsiteScanResult): Promise<string[]> {
+        let processedUrls = discoveredUrls;
+        if (isNil(discoveredUrls)) {
+            return [];
+        }
+
+        if (!isNil(websiteScanResult) && !isNil(websiteScanResult.knownPages)) {
```
RobGallo, line 19: Would optional chaining work here? (i.e., the ?. operator) It would take this fown from one condition to two.

packages/web-api-scan-runner/src/crawler/discovered-urls-processor.ts
```diff
@@ -0,0 +1,41 @@
...
+            processedUrls = this.removeUrlsFromList(discoveredUrls, websiteScanResult.knownPages);
+        }
+        processedUrls = this.limitNumUrls(processedUrls, await this.getUrlLimit(websiteScanResult));
```
RobGallo, line 22: I'm not wild about the complexity of the line here. But more importantly, we should only ever need to call getUrlLimit once. We should probably pass it into the class as a parameter for example.

packages/web-api-scan-runner/src/crawler/discovered-urls-processor.ts
```diff
@@ -0,0 +1,41 @@
...
+    private removeUrlsFromList(urlList: string[], removeUrls: string[]): string[] {
+        return _.pullAll(urlList, removeUrls);
+    }
+
+    private limitNumUrls(urlList: string[], urlLimit: number): string[] {
+        return _.take(urlList, urlLimit);
+    }
+
+    private async getUrlLimit(websiteScanResult?: WebsiteScanResult): Promise<number> {
+        const urlCrawlLimit = (await this.serviceConfig.getConfigValue('crawlConfig')).urlCrawlLimit;
```
RobGallo, line 36: Is it ever possible for the crawl limit to be a negative number?

pownkel: We configure the crawl limit ourselves, so while it's possible for it to be set to a negative number, in practice that's not something we'd ever expect to see

packages/web-api-scan-runner/src/crawler/discovered-urls-processor.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    });
+
+    it('handles missing knownPages', () => {
+        const processedUrls = testSubject.getProcessedUrls(urlsList, maxUrlsLimit, {} as WebsiteScanResult);
+
+        expect(processedUrls).toEqual(urlsList);
+    });
+
+    it('handles knownUrls.length > urlCrawlLimit', () => {
+        const urlCrawlLimit = knownUrls.length - 1;
```
RobGallo, line 60: Is "- 1" necessary? If there are 100 known urls and the limit is 100, we would still expect no urls to be added, correct?

packages/web-api-scan-runner/src/crawler/discovered-urls-processor.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { injectable } from 'inversify';
+import _, { isNil } from 'lodash';
+import { WebsiteScanResult } from 'storage-documents';
+
+@injectable()
+export class DiscoveredUrlsProcessor {
```
RobGallo, line 9: It seems like this doesn't need to be a class since it maintains no state data. We could just export the main function.

packages/web-api-scan-runner/src/crawler/discovered-urls-processor.ts
```diff
@@ -0,0 +1,37 @@
...
+    public getProcessedUrls(discoveredUrls: string[], urlCrawlLimit: number, websiteScanResult?: WebsiteScanResult): string[] {
+        let processedUrls = discoveredUrls;
+        if (isNil(discoveredUrls)) {
+            return [];
+        }
+
+        if (!isNil(websiteScanResult?.knownPages)) {
+            processedUrls = this.removeUrlsFromList(discoveredUrls, websiteScanResult.knownPages);
+        }
+        processedUrls = this.limitNumUrls(processedUrls, this.getUrlLimit(urlCrawlLimit, websiteScanResult));
```
RobGallo, line 19: You can get rid of the assignment here and just return the value of limitNumUrls.

lamaks (dismissed):

packages/web-api-scan-runner/src/crawler/discovered-urls-processor.spec.ts
pownkel: I wanted to test the case where urlCrawlLimit - knownUrls.length is negative to make sure it didn't throw any errors

RobGallo (approved):

PR #1250, opened 2021-01-22
feat(cloud-crawler): Create CrawlRunner class to manage crawling during a run

RobGallo:
#### Description of changes

At present, this class manages logging and error handling for crawling during a run. In future, it can also manage adding known urls to scan results.

Note: In a few places, the use of `inversify.Container` was changed to `inversify.interfaces.Container` because functions which create the crawler only have the interface type available.

#### Pull request checklist

### DIFF
--- a/packages/web-api-scan-runner/src/crawl-runner/crawl-runner.spec.ts
+++ b/packages/web-api-scan-runner/src/crawl-runner/crawl-runner.spec.ts
@@ -7,6 +7,7 @@ import { Crawler, SimpleCrawlerRunOptions } from 'accessibility-insights-crawler
 import { GlobalLogger } from 'logger';
 import { Page } from 'puppeteer';
 import { IMock, It, Mock, MockBehavior } from 'typemoq';
+import { ServiceConfiguration } from 'common';
 import { ScanMetadataConfig } from '../scan-metadata-config';
 import { ScanMetadata } from '../types/scan-metadata';
 import { CrawlRunner } from './crawl-runner';
@@ -26,9 +27,11 @@ describe('CrawlRunner', () => {
     const baseUrl = 'testUrl';
     const discoveryPatterns = ['testPattern'];
     const page = {} as Page;
+    const urlCrawlLimit = 42;
 
     let loggerMock: IMock<GlobalLogger>;
     let scanMetaDataConfigMock: IMock<ScanMetadataConfig>;
+    let serviceConfigMock: IMock<ServiceConfiguration>;
 
     beforeEach(() => {
         loggerMock = Mock.ofType<GlobalLogger>(undefined, MockBehavior.Loose);
@@ -39,17 +42,25 @@ describe('CrawlRunner', () => {
             .setup((m) => m.getConfig())
             .returns(() => ({ id: scanId } as ScanMetadata))
             .verifiable();
+
+        serviceConfigMock = Mock.ofType<ServiceConfiguration>(undefined, MockBehavior.Strict);
     });
 
     afterEach(() => {
         loggerMock.verifyAll();
         scanMetaDataConfigMock.verifyAll();
+        serviceConfigMock.verifyAll();
     });
 
     it('returns undefined when crawler provider returns null', async () => {
         const crawlerProviderMock = createCrawlerProviderMock(null);
 
-        const crawlRunner = new CrawlRunner(crawlerProviderMock.object, loggerMock.object, scanMetaDataConfigMock.object);
+        const crawlRunner = new CrawlRunner(
+            crawlerProviderMock.object,
+            loggerMock.object,
+            serviceConfigMock.object,
+            scanMetaDataConfigMock.object,
+        );
 
         expect(await crawlRunner.run(baseUrl, discoveryPatterns, page)).toBeUndefined();
 
@@ -65,7 +76,14 @@ describe('CrawlRunner', () => {
 
         const crawlerProviderMock = createCrawlerProviderMock(crawlerMock.object);
 
-        const crawlRunner = new CrawlRunner(crawlerProviderMock.object, loggerMock.object, scanMetaDataConfigMock.object);
+        initServiceConfigMock();
+
+        const crawlRunner = new CrawlRunner(
+            crawlerProviderMock.object,
+            loggerMock.object,
+            serviceConfigMock.object,
+            scanMetaDataConfigMock.object,
+        );
 
         const retVal = await crawlRunner.run(baseUrl, discoveryPatterns, page);
 
@@ -80,6 +98,7 @@ describe('CrawlRunner', () => {
             baseUrl,
             discoveryPatterns,
             page,
+            maxRequestsPerCrawl: urlCrawlLimit,
         } as SimpleCrawlerRunOptions;
 
         let actualRunOptions: SimpleCrawlerRunOptions;
@@ -94,7 +113,14 @@ describe('CrawlRunner', () => {
 
         const crawlerProviderMock = createCrawlerProviderMock(crawlerMock.object);
 
-        const crawlRunner = new CrawlRunner(crawlerProviderMock.object, loggerMock.object, scanMetaDataConfigMock.object);
+        initServiceConfigMock();
+
+        const crawlRunner = new CrawlRunner(
+            crawlerProviderMock.object,
+            loggerMock.object,
+            serviceConfigMock.object,
+            scanMetaDataConfigMock.object,
+        );
 
         const actualRetVal = await crawlRunner.run(baseUrl, discoveryPatterns, page);
 
@@ -113,5 +139,14 @@ describe('CrawlRunner', () => {
             .verifiable();
 
         return crawlerProviderMock;
+    }
+
+    function initServiceConfigMock(): void {
+        serviceConfigMock
+            .setup((m) => m.getConfigValue('crawlConfig'))
+            .returns(async () => ({
+                urlCrawlLimit,
+            }))
+            .verifiable();
     }
 });
--- a/packages/web-api-scan-runner/src/crawl-runner/crawl-runner.spec.ts
+++ b/packages/web-api-scan-runner/src/crawl-runner/crawl-runner.spec.ts
@@ -32,10 +32,13 @@ describe('CrawlRunner', () => {
 
     beforeEach(() => {
         loggerMock = Mock.ofType<GlobalLogger>(undefined, MockBehavior.Loose);
-        loggerMock.setup((m) => m.setCommonProperties({ scanId: scanId }));
+        loggerMock.setup((m) => m.setCommonProperties({ scanId: scanId })).verifiable();
 
         scanMetaDataConfigMock = Mock.ofType<ScanMetadataConfig>(undefined, MockBehavior.Strict);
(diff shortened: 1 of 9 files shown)


packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
```diff
@@ -0,0 +1,54 @@
...
+
+@injectable()
+export class CrawlRunner {
+    constructor(
+        @inject(crawlerIocTypes.CrawlerProvider) private readonly getCrawler: CrawlerProvider,
+        @inject(GlobalLogger) private readonly logger: GlobalLogger,
+        @inject(ScanMetadataConfig) scanMetadataConfig: ScanMetadataConfig,
+    ) {
+        const scanMetadata = scanMetadataConfig.getConfig();
+        this.logger.setCommonProperties({ scanId: scanMetadata.id });
```
pownkel, line 20: We shouldn't need to set logger properties in this class. GlobalLogger is a singleton per batch worker task, and setCommonProperties is already called in Runner.run(), so the common logger properties will already be set by the time CrawlRunner.run() executes.

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
```diff
@@ -0,0 +1,54 @@
...
+        } as SimpleCrawlerRunOptions;
+
+        this.logger.logInfo('begin crawl');
+
+        let retVal: string[] | undefined;
+
+        try {
+            retVal = await crawler.crawl(crawlerRunOptions);
+        } catch (ex) {
+            this.logger.logError('Exception thrown during crawl');
```
pownkel, line 44: Instead of logging the exception on a separate line below, could we do:
```suggestion
            this.logger.logError('Exception thrown during crawl', { error: JSON.stringify(ex) });
```

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
```diff
@@ -0,0 +1,54 @@
...
+            this.logger.logInfo(ex);
+
+            return undefined;
+        }
+
+        this.logger.logInfo('crawl succeeded');
```
pownkel, line 50: Could you add something to this log line about how many urls were discovered?

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
```diff
@@ -0,0 +1,54 @@
...
+        if (crawler == null) {
+            this.logger.logInfo('No crawler provided by crawler provider');
+
+            return undefined;
+        }
+
+        const crawlerRunOptions = {
+            baseUrl,
+            discoveryPatterns,
+            page,
```
pownkel, line 34: It might be a good idea to add a value for maxRequestsPerCrawl here (probably the value of the configured urlCrawlLimit)

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+    });
+
+    it('returns undefined when crawler provider returns null', async () => {
+        const crawlerProviderMock = createCrawlerProviderMock(null);
+
+        const crawlRunner = new CrawlRunner(crawlerProviderMock.object, loggerMock.object, scanMetaDataConfigMock.object);
+
+        expect(await crawlRunner.run(baseUrl, discoveryPatterns, page)).toBeUndefined();
+
+        crawlerProviderMock.verifyAll();
```
pownkel, line 53: I don't see any verifiable calls set up for this mock, should the call set up in createCrawlerProviderMock be set as verifiable?

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
```diff
@@ -0,0 +1,54 @@
...
+    public async run(baseUrl: string, discoveryPatterns: string[], page: Page): Promise<string[] | undefined> {
+        const crawler = await this.getCrawler();
+        if (crawler == null) {
+            this.logger.logInfo('No crawler provided by crawler provider');
+
+            return undefined;
+        }
+
+        const crawlerRunOptions = {
+            baseUrl,
```
lamaks, line 32: The options set is confusing here b/c we do plan to scan a single or set of pages like five pages at the time. In this case we should pass discovery pattern and array pages to crawl. The base URL is only need if we start continues crawl from base URL and to define default discovery pattern

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
```diff
@@ -0,0 +1,54 @@
...
+            discoveryPatterns,
+            page,
+        } as SimpleCrawlerRunOptions;
+
+        this.logger.logInfo('begin crawl');
```
lamaks, line 37:
```suggestion
        this.logger.logInfo('Starting web page crawling');
```

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
lamaks, line 50:
```suggestion
        this.logger.logInfo('Web page crawling completed successfully');
```

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
lamaks, line 44:
```suggestion
            this.logger.logError('Failure while crawling web page');
```

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { GlobalLogger } from 'logger';
+import { inject, injectable } from 'inversify';
+import { Crawler, iocTypes as crawlerIocTypes, SimpleCrawlerRunOptions } from 'accessibility-insights-crawler';
+import { Page } from 'puppeteer';
+import { ScanMetadataConfig } from '../scan-metadata-config';
+
```
lamaks, line 9:
```suggestion
import { System } from 'common';
```

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.spec.ts
```diff
@@ -0,0 +1,107 @@
...
+        } as SimpleCrawlerRunOptions;
+
+        let actualRunOptions: SimpleCrawlerRunOptions;
+        const expectedRetVal = ['discoveredUrl'];
+
+        const crawlerMock = Mock.ofType(CrawlerMock);
+        crawlerMock
+            .setup((m) => m.crawl(It.isAny()))
+            .callback((runOptions) => (actualRunOptions = runOptions))
+            .returns(async () => expectedRetVal);
```
pownkel, line 86: Can be simplified to
```suggestion
            .setup((m) => m.crawl(actualRunOptions))
            .returns(async () => expectedRetVal)
            .verifiable();
```
This would also remove the need to validate the value of actualRunOptions at the end of the test

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
pownkel: When we scan multiple pages at a time, where do those urls come from? We assumed that each time the crawler is called in the batch worker we're only discovering URLs on a single page, which is the same page that is scanned in the same batch task.

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
pownkel, line 34:
```suggestion
            page.getUnderlyingPage(),
```
Uses the puppeteer page instead of our wrapper

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.spec.ts
RobGallo: Mocks are verifiable by default.

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
RobGallo: I've applied this suggestion in combination with Kelsey's suggestion.

RobGallo: I've applied this suggestion in combination with Maxim's suggestion.

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.spec.ts
RobGallo: I mistakenly thought that it wasn't necessary to call verifiable. I conflated that with the fact that it isn't necessary to pass a parameter to verifiable if you intent the function to be called only once.

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.spec.ts
RobGallo: I don't think so. The parameter indicates an object to match, not an object to assign the passed in value to. Plus I tried it just to be sure and I got an error saying actualRunOptions can't be undefined, which it is at that point.

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
RobGallo: There's no guarantee the id will be set at the point when the CrawlRunner may need it. The Runner class sets the id, but because the CrawlRunner is actually instantiated first, the CrawlRunner is now the first object to set the id.
This is a bit of a flaw in the design IMO. The logger interface consumed by classes which need to log should not give access to setting the scan id if they aren't intended to set it.
In the end, for the moment, we know that the logger isn't used in the CrawlRunner before the scan id is set. But that would be very easy to overlook if the code were to change.
My current thinking is that I'd rather set the id twice than cause a problem if the code changes. I also think the problem wouldn't be detected by unit tests if the code changes.

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
RobGallo: Done.

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.spec.ts
pownkel: Oops, I meant expectedRunOptions, not actualRunOptions. That will match against the expected value without the additional equality check below.

RobGallo: So I didn't do this initially because I wasn't certain the equality comparison between two objects in the mock call would work. Turns out it does, so I've made the change. I will say this though: it makes the error message you get when the two objects don't match a good bit more confusing.

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
RobGallo: Done.

pownkel (approved):

PR #1252, opened 2021-01-22
refactor(cloud-crawler): refactor out axe result merger of runner

waabid:
#### Description of changes

I modified the logic slightly as there was a bit of redundancy in the checks.

#### Pull request checklist

### DIFF
--- a/packages/web-api-scan-runner/src/runner/axe-result-merger.ts
+++ b/packages/web-api-scan-runner/src/runner/axe-result-merger.ts
@@ -3,11 +3,12 @@
 
 import axe from 'axe-core';
 import { AxeResultsReducer } from 'axe-result-converter';
-import { inject } from 'inversify';
+import { inject, injectable } from 'inversify';
 import { GlobalLogger } from 'logger';
 import { CombinedScanResultsProvider, CombinedScanResultsReadResponse } from 'service-library';
 import { CombinedScanResults } from 'storage-documents';
 
+@injectable()
 export class AxeResultMerger {
     public constructor(
         @inject(GlobalLogger) private readonly logger: GlobalLogger,
--- a/packages/web-api-scan-runner/src/runner/axe-result-merger.ts
+++ b/packages/web-api-scan-runner/src/runner/axe-result-merger.ts
@@ -5,8 +5,9 @@ import axe from 'axe-core';
 import { AxeResultsReducer } from 'axe-result-converter';
 import { inject, injectable } from 'inversify';
 import { GlobalLogger } from 'logger';
-import { CombinedScanResultsProvider, CombinedScanResultsReadResponse } from 'service-library';
+import { CombinedScanResultsProvider } from 'service-library';
 import { CombinedScanResults } from 'storage-documents';
+import { CombinedResultsBlobInfo } from './combined-results-blob-getter';
 
 @injectable()
 export class AxeResultMerger {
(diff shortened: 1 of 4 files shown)


packages/web-api-scan-runner/src/runner/axe-result-merger.ts
```diff
@@ -0,0 +1,81 @@
...
+        combinedScanResults.urlCount.total++;
+        if (axeScanResults.violations.length > 0) {
+            combinedScanResults.urlCount.failed++;
+        } else {
+            combinedScanResults.urlCount.passed++;
+        }
+
+        this.axeResultsReducer.reduce(combinedScanResults.axeResults, axeScanResults);
+        const blobWriteResponse = await this.combinedScanResultsProvider.writeCombinedResults(
+            combinedResultsBlobId,
```
pownkel, line 32: if an undefined blob Id is passed in, we'll need to generate a new one somewhere before writing (and return the new blob id so it can be updated in the WebsiteScanResult document)

packages/web-api-scan-runner/src/runner/axe-result-merger.ts
```diff
@@ -0,0 +1,81 @@
...
+@injectable()
+export class AxeResultMerger {
+    public constructor(
+        @inject(GlobalLogger) private readonly logger: GlobalLogger,
+        @inject(CombinedScanResultsProvider) protected readonly combinedScanResultsProvider: CombinedScanResultsProvider,
+        @inject(AxeResultsReducer) protected readonly axeResultsReducer: AxeResultsReducer,
+    ) {}
+
+    public async mergeAxeResults(axeScanResults: axe.AxeResults, combinedResultsBlobId: string): Promise<CombinedScanResults> {
+        const blobReadResponse = await this.getOrCreateCombinedResultsBlob(combinedResultsBlobId);
```
lamaks, line 20: This is not redundant logic. We have two blob ids in runner code flow: 1st from websiteScanResult, second is generated if websiteScanResult has no blob id yet:  combinedResultsBlobId. In case websiteScanResult has no id we do create new blob body w/o accessing blob storage hence we have createNew option. The combinedResultsBlobId will always have value: `        const combinedResultsBlobId = websiteScanResult.combinedResultsBlobId ?? this.guidGenerator.createGuid();
`

`    private async mergeAxeResults(
        axeScanResults: axe.AxeResults,
        combinedResultsBlobId: string,
        createNew: boolean,
    ): Promise<CombinedScanResults> {
        const blobReadResponse = await this.getOrCreateCombinedResultsBlob(createNew ? undefined : combinedResultsBlobId);
`

packages/web-api-scan-runner/src/runner/combined-results-blob-getter.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+            blobId: givenResultsBlobIdStub,
+            response: readResponseStub,
+        };
+
+        setupBlobRead(givenResultsBlobIdStub, readResponseStub);
+
+        expect(await testSubject.getBlobInfo(givenResultsBlobIdStub)).toEqual(expectedBlobInfo);
+    });
+
+    test('readResponse has blobNotFound error code', async () => {
```
pownkel, line 80:
```suggestion
    test('readResponse has jsonParseError error code', async () => {
```

pownkel (approved):

PR #1271, opened 2021-01-28
feat(cloud-crawler): Run crawler and save results

pownkel:
#### Description of changes

- create DeepScanner to coordinate reading crawl data, crawling, and saving the results (will be expanded in a future feature to include creating and queueing additional scan requests)
- run deep scan if the original request has `deepScan = true`
- Deduplicate discoveryPatterns when saving websiteScanResult
- Renamed iocTypes in the crawler container to crawlerIocTypes

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1793075


packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
```diff
@@ -19,14 +18,12 @@ export class CrawlRunner {
         @inject(crawlerIocTypes.CrawlerProvider) private readonly getCrawler: CrawlerProvider,
         @inject(GlobalLogger) private readonly logger: GlobalLogger,
         @inject(ServiceConfiguration) private readonly serviceConfig: ServiceConfiguration,
-        @inject(ScanMetadataConfig) scanMetadataConfig: ScanMetadataConfig,
         @inject(BatchConfig) private readonly batchConfig: BatchConfig,
-    ) {
-        const scanMetadata = scanMetadataConfig.getConfig();
-        this.logger.setCommonProperties({ scanId: scanMetadata.id });
```
pownkel, line 26: Moved this into run() because it was causing our dependency injection tests to fail (we expect logger.setup() to be called before any other logger methods, so calling logger methods in the constructor of an auto-injected object makes it more complicated to guarantee that)

packages/web-api-scan-runner/src/crawl-runner/process-discovered-urls.ts
```diff
@@ -2,19 +2,20 @@
 // Licensed under the MIT License.
 
 import _, { isNil } from 'lodash';
-import { WebsiteScanResult } from 'storage-documents';
 
-export function processDiscoveredUrls(discoveredUrls: string[], urlCrawlLimit: number, websiteScanResult?: WebsiteScanResult): string[] {
+export type DiscoveredUrlProcessor = typeof processDiscoveredUrls;
+
+export function processDiscoveredUrls(discoveredUrls: string[], urlCrawlLimit: number, knownUrls?: string[]): string[] {
```
lamaks, line 8: may be rename to `discoveredUrlProcessor`

packages/web-api-scan-runner/src/crawl-runner/deep-scanner.ts
```diff
@@ -0,0 +1,50 @@
...
+
+@injectable()
+export class DeepScanner {
+    constructor(
+        @inject(GlobalLogger) private readonly logger: GlobalLogger,
+        @inject(CrawlRunner) private readonly crawlRunner: CrawlRunner,
+        @inject(WebsiteScanResultProvider) private readonly websiteScanResultProvider: WebsiteScanResultProvider,
+        @inject(ServiceConfiguration) private readonly serviceConfig: ServiceConfiguration,
+        @inject(WebsiteScanResultUpdater) private readonly websiteScanResultUpdater: WebsiteScanResultUpdater,
+        private readonly processUrls: DiscoveredUrlProcessor = processDiscoveredUrls,
```
lamaks, line 25: may be rename to `discoveredUrlProcessor`

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
RobGallo, line 26: You've removed this line, but didn't replace it. There's still no guarantee this will be set before it's required.

packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -77,6 +80,9 @@ export class Runner {
         try {
             await this.openPage(scanMetadata.url);
             axeScanResults = await this.scan(pageScanResult, this.page);
+
+            await this.deepScan(scanMetadata, pageScanResult);
```
lamaks, line 84: We are not ready to use crawler from runner. The crawler form cli does crawl **and** scan. What we need here is crawl only. The arch can look like crawler crawl and scan using runner impl similar to cli impl when scan processor is invoked from crawler.

pownkel: The crawler that's registered to the web-api-scan-runner container uses SimpleCrawlerEngine, which shares an interface with the cli CrawlerEngine, but only collects links from one web page without any scanning.

packages/web-api-scan-runner/src/crawl-runner/crawl-runner.ts
```diff
@@ -19,14 +18,12 @@ export class CrawlRunner {
         @inject(crawlerIocTypes.CrawlerProvider) private readonly getCrawler: CrawlerProvider,
```
lamaks, line 18: Confusing name alias. If `iocTypes` is exported from cli it should be named as `crawlerIocTypes` in cli instead of aliasing in `import`

packages/web-api-scan-runner/src/runner/runner.ts
lamaks: Still this is not expected pattern. The `SimpleCrawlerEngine` crawler will create browser instance, make call to the web page, render the page, and crawl the page. All this duplicate runner except getting page links. We need to remove scanner inclusion from runner for now.

lamaks (approved):

PR #1272, opened 2021-01-29
refactor: ScanRunnerTelemetryManager

pownkel:
#### Description of changes

Extract the trackEvent calls in runner.ts to another class, ScanRunnerTelemetryManager. The new class is also responsible for managing the "scan started" and "scan completed" timestamps. This simplifies the Runner unit tests, since it a) removes the need to mock Date.now() calls in runner tests, and b) reduces the number of logger calls to be verified.

#### Pull request checklist

### DIFF
--- a/packages/web-api-scan-runner/src/runner/runner.spec.ts
+++ b/packages/web-api-scan-runner/src/runner/runner.spec.ts
@@ -232,9 +232,9 @@ describe(Runner, () => {
 
         setupUpdateScanRunResultCall(getFailingJobStateScanResult(unscannableAxeScanResults.error));
 
-        telemetryManagerMock.setup(t => t.trackScanStarted(scanSubmittedDate)).verifiable();
-        telemetryManagerMock.setup(t => t.trackScanTaskFailed()).verifiable(Times.never());
-        telemetryManagerMock.setup(t => t.trackScanCompleted()).verifiable();
+        telemetryManagerMock.setup((t) => t.trackScanStarted(scanSubmittedDate)).verifiable();
+        telemetryManagerMock.setup((t) => t.trackScanTaskFailed()).verifiable(Times.never());
+        telemetryManagerMock.setup((t) => t.trackScanCompleted()).verifiable();
 
         await runner.run();
     });
@@ -246,9 +246,9 @@ describe(Runner, () => {
 
         setupUpdateScanRunResultCall(getFailingJobStateScanResult(System.serializeError(failureMessage), false));
 
-        telemetryManagerMock.setup(t => t.trackScanStarted(scanSubmittedDate)).verifiable();
-        telemetryManagerMock.setup(t => t.trackScanTaskFailed()).verifiable();
-        telemetryManagerMock.setup(t => t.trackScanCompleted()).verifiable();
+        telemetryManagerMock.setup((t) => t.trackScanStarted(scanSubmittedDate)).verifiable();
+        telemetryManagerMock.setup((t) => t.trackScanTaskFailed()).verifiable();
+        telemetryManagerMock.setup((t) => t.trackScanCompleted()).verifiable();
 
         await runner.run();
     });
@@ -263,7 +263,7 @@ describe(Runner, () => {
                 ),
             )
             .verifiable();
-        telemetryManagerMock.setup(t => t.trackScanStarted(scanSubmittedDate)).verifiable(Times.never());
+        telemetryManagerMock.setup((t) => t.trackScanStarted(scanSubmittedDate)).verifiable(Times.never());
 
         await runner.run();
     });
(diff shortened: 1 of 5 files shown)


packages/web-api-scan-runner/src/runner/runner.spec.ts
```diff
@@ -240,6 +232,10 @@ describe(Runner, () => {
 
         setupUpdateScanRunResultCall(getFailingJobStateScanResult(unscannableAxeScanResults.error));
 
+        telemetryManagerMock.setup((t) => t.trackScanStarted(scanSubmittedDate)).verifiable();
+        telemetryManagerMock.setup((t) => t.trackScanTaskFailed()).verifiable(Times.never());
```
RobGallo, line 236: Have you considered using Mock.Behavior.Strict when this object is created? That way, TypeMoq will throw an error whenever a function that isn't mocked is called. I use it whenever I can because it is easy to forget that a certain function isn't expected to be called, (i.e., it's more future proof) There are times when you can't use it though, and that's reasonable too.

packages/web-api-scan-runner/src/runner/runner.ts
```diff
@@ -67,14 +69,8 @@ export class Runner {
...
 
-        const scanStartedTimestamp: number = Date.now();
-        const scanSubmittedTimestamp: number = this.guidGenerator.getGuidTimestamp(scanMetadata.id).getTime();
-
-        this.logger.trackEvent('ScanRequestRunning', undefined, { runningScanRequests: 1 });
-        this.logger.trackEvent('ScanTaskStarted', undefined, {
-            scanWaitTime: (scanStartedTimestamp - scanSubmittedTimestamp) / 1000,
-            startedScanTasks: 1,
-        });
+        const scanSubmittedTimestamp = this.guidGenerator.getGuidTimestamp(scanMetadata.id);
```
RobGallo, line 72: Is there a reason creating the timestamp didn't get brought into the telemetry manager?

RobGallo, line 72: Could this be moved into ScanRunnerTelemetryManager?

packages/web-api-scan-runner/src/scan-runner-telemetry-manager.ts
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import { inject, injectable } from 'inversify';
+import { isNil } from 'lodash';
+import { GlobalLogger, ScanTaskCompletedMeasurements } from 'logger';
+
+@injectable()
+export class ScanRunnerTelemetryManager {
+    public scanSubmitted: number;
```
RobGallo, line 10: These probably shouldn't be public because their only public use is unit testing. The problem is essentially one of reading the code: When I first read this class I automatically assumed that because these properties are public, they are required by other production code.
One way you could fix this is to make the values protected; and in the test file create a class which inherits from ScanRunnerTelemetryManager and exposes access to them. (That may not be the only way option.)

packages/web-api-scan-runner/src/scan-runner-telemetry-manager.ts
```diff
@@ -0,0 +1,50 @@
...
+        this.logger.trackEvent('ScanTaskFailed', undefined, { failedScanTasks: 1 });
+    }
+
+    public trackScanCompleted(): void {
+        if (isNil(this.scanStarted) || isNil(this.scanSubmitted)) {
+            return;
+        }
+        const scanCompletedTimestamp: number = this.getCurrentTimestamp();
+        const telemetryMeasurements: ScanTaskCompletedMeasurements = {
+            scanExecutionTime: (scanCompletedTimestamp - this.scanStarted) / 1000,
```
RobGallo, line 43: Knit: There are a few places in this file where you divide by the number 1000. Without looking at the documentation for TimeStamp, I'm not sure whether you're calculating in milliseconds, microseconds, or nanoseconds. It would be faster to comprehend and require no long-term memory of the TimeStamp class if you created a simple function called something like `inMilliseconds`.

packages/web-api-scan-runner/src/scan-runner-telemetry-manager.spec.ts
```diff
@@ -15,14 +26,17 @@ describe(ScanRunnerTelemetryManager, () => {
...
     let getCurrentDateMock: IMock<() => number>;
+    let guidGeneratorMock: IMock<GuidGenerator>;
 
-    let testSubject: ScanRunnerTelemetryManager;
+    let testSubject: TestableScanRunnerTelemetryManager;
 
     beforeEach(() => {
         loggerMock = Mock.ofType<Logger>();
         getCurrentDateMock = Mock.ofInstance(() => null);
+        guidGeneratorMock = Mock.ofType<GuidGenerator>();
```
RobGallo, line 36: Should this also use MockBehavior.Strict?

RobGallo (approved):

PR #1287, opened 2021-02-11
feat(host-browser-service): update webdriver to accept ws-endpoint for puppeteer connection

karanbirsingh:
#### Details

This PR adds `puppeteer.connect` functionality to inner `webdriver` and `page` classes.

`puppeteer-options` is moved into its own file so that the `host-browser-service` can reuse it eventually.

The coniditional `launch/connect` decision is in `page` for now.

##### Motivation

Enable `web-api-scan-runner` to run scans via `puppeteer.connect` in future PRs

##### Context

This PR only adds new methods & updates existing usages - production behavior should not be changed in this PR. All scans still use `launch`, not `connect` yet.

#### Pull request checklist

### DIFF
--- a/packages/scanner-global-library/src/page.ts
+++ b/packages/scanner-global-library/src/page.ts
@@ -12,6 +12,11 @@ import { WebDriver } from './web-driver';
 import { PageNavigator } from './page-navigator';
 import { BrowserError } from './browser-error';
 
+export interface BrowserStartOptions {
+    browserExecutablePath?: string;
+    browserWSEndpoint?: string;
+}
+
 @injectable()
 export class Page {
     public requestUrl: string;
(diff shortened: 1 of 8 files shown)


packages/scanner-global-library/src/page.ts
```diff
@@ -33,8 +38,13 @@ export class Page {
         @inject(GlobalLogger) @optional() private readonly logger: GlobalLogger,
     ) {}
 
-    public async create(browserExecutablePath?: string): Promise<void> {
-        this.browser = await this.webDriver.launch(browserExecutablePath);
+    public async create(options?: BrowserStartOptions): Promise<void> {
+        if (options?.browserWSEndpoint !== undefined) {
+            this.browser = await this.webDriver.connect(options.browserWSEndpoint);
```
AhmedAbdoOrtiga, line 43: can we do the fall back here to the browser executable if it fails to launch

AhmedAbdoOrtiga (approved):

packages/scanner-global-library/src/page.ts
karanbirsingh: discussed offline, we plan to have the fallback logic in the web-api-scan-runner so each scan entry point can make the decision independently

PR #1305, opened 2021-02-26
chore(deps): upgrade url-parse for security fix

lisli1:
#### Details

This PR updates the transitive dependency url-parse:
1. Removed url-parse from yarn.lock
2. Ran `yarn`
3. Verified that url-parse in yarn.lock points to its latest version, 1.5.1

##### Motivation

Addresses [CVE-2021-27515](https://nvd.nist.gov/vuln/detail/CVE-2021-27515): url-parse before 1.5.0 mishandles certain uses of backslash such as http:\/ and interprets the URI as a relative path.

##### Context

Did not update other dependencies of root dependency `@azure/batch`

#### Pull request checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -12612,9 +12612,9 @@ urix@^0.1.0:
   integrity sha1-2pN/emLiH+wf0Y1Js1wpNQZ6bHI=
 
 url-parse@^1.4.7:
-  version "1.4.7"
-  resolved "https://registry.yarnpkg.com/url-parse/-/url-parse-1.4.7.tgz#a8a83535e8c00a316e403a5db4ac1b9b853ae278"
-  integrity sha512-d3uaVyzDB9tQoSXFvuSUNFibTd9zxd2bkVrDRvF5TmvWWQwqE4lgYJ5m+x1DbecWkw+LK4RNl2CU1hHuOKPVlg==
+  version "1.5.1"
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -62,6 +62,14 @@
     "@opentelemetry/api" "^0.6.1"
     tslib "^2.0.0"
 
+"@azure/core-auth@^1.1.4":
```
lamaks, line 65: Is there only way to edit yarn.lock for version update?

lisli1: I can revert my last commit and update only url-parse in yarn.lock (first commit)

lamaks: will the `resolutions` under root\package.json work instead? I am concerning about editing yarn.lock, usually if we run `yarn install` it gets reverted to the original state.

lisli1: `resolutions` should not be necessary in this case, and ideally they are not used unless absolutely necessary. The (only) root dependency for this component is `@azure/batch`, which lists `url-parse "^1.4.7"` with caret `^` version, so any minor releases are acceptable, including 1.5.1. I also edited yarn.lock by removing the entry for url-parse and then running `yarn install`, which re-added url-parse with its newest acceptable version. Since I ran `yarn install` to get this update, running `yarn install` in the future will not revert it to 1.4.7 unless `@azure/batch` changes its dependency to be ~1.4.7 (patch releases acceptable), 1.4.7 (specifically this version), or similar.

lamaks (approved):

PR #1311, opened 2021-03-02
feat(host-browser-service): update browser-server to launch browsers upon puppeteer.connect()

karanbirsingh:
#### Details

This PR updates `browser-server` to accept requests from `puppeter.connect({ wsBrowserEndpoint: '<host>:8585'})`. The request handler [listens for `upgrade` events](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism#upgrading_to_a_websocket_connection) to intercept expected websocket connections; the server proxies the request to a freshly launched browser (similar to [here](https://github.com/websockets/ws#multiple-servers-sharing-a-single-https-server)).

The original approach [here](https://github.com/karanbirsingh/axe-docker-remote-browser/blob/main/start-browser-server.js) with explicit connections to a returned `wsEndpoint` didn't work in testing. When remoted into our Azure Batch node, puppeteer clients inside the Windows Docker container were able to make requests to the `localhost:8585` server, but were consistently met with `connect ETIMEDOUT at TCPConnectWrap.afterConnect` when connecting to a host URL of the form `ws://<returned-wsEndpoint-url>/devtools...`. When testing the proxy instead, clients inside the container were able to successfully drive scans on the host. One side effect of this change is that clients no longer need to retrieve individual endpoint URLs; they all share a common URL.

sample client code used in tests:
```js
const browser = await puppeteer.connect({
    browserWSEndpoint: `ws://${host_ip}:8585`,
});
```

This PR also includes some initial error handling & a new file `browser-launcher`, which encapsulates launch/close behavior & keeps track of launched browsers.

##### Motivation

This PR doesn't change any client code yet. It adds to the host code so we can simplify our testing in separate resource groups.

##### Context

out of scope for this PR:
- closing stale browsers
- potentially wrapping in an OS (windows) service
- changes to `web-api-scan-runner` to actually use remote browsers
- startup script to install dependencies on the host & run the server

##### Testing

On my local machine, I tested this by running `node host-browser-service.js` (output of `yarn build`) with an app insights instrumentation key set in environment variables. I then wrote a client script that runs an axe-puppeteer scan via `browserWSEndpoint: 'ws://localhost:8585'`. I ran several instances of this script to ensure I could see scan results. I terminated several client scripts early to check incoming socket errors on the server.

When remoted into a resource group, I ran a simpler version of this proxy from Powershell alongside a version of the client script inside a Windows container. The special identifier `host.docker.internal` [from the prototype](https://github.com/karanbirsingh/axe-docker-remote-browser/blob/7a932abef4d14805bae676555d1a2aadd889cbd1/index.js#L14) wasn't available in this environment, so we had to inline the host IP explicitly. I was able to see scan results executed on the host.

Neither of these approaches used Batch-initiated docker tasks, so we'll be testing that when we update client code.

#### Pull request checklist


pownkel (approved):
LGTM

PR #1317, opened 2021-03-05
Chore: Reduce Job Manager Task Restart Time

AhmedAbdoOrtiga:
#### Details
The job manager task timesout sometimes before hitting the code that check for that, and that causes some scans to be pending forever since some of the tasks will be terminated before even starting or while running.
So in this PR we will give some margin to the job manager task and pending runner tasks to end before the max clock wall time expires.

#### Pull request checklist

### DIFF
--- a/packages/service-library/src/batch/batch-task-creator.ts
+++ b/packages/service-library/src/batch/batch-task-creator.ts
@@ -45,7 +45,7 @@ export abstract class BatchTaskCreator {
         }
 
         this.activeScanMessages = [];
-        const restartAfterTime = moment().add(this.jobManagerConfig.maxWallClockTimeInHours, 'hour').toDate();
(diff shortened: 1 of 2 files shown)


packages/service-library/src/batch/batch-task-creator.ts
```diff
@@ -45,7 +45,9 @@ export abstract class BatchTaskCreator {
         }
 
         this.activeScanMessages = [];
-        const restartAfterTime = moment().add(this.jobManagerConfig.maxWallClockTimeInHours, 'hour').toDate();
+        const restartAfterTime = moment()
+            .add(this.jobManagerConfig.maxWallClockTimeInHours / 2, 'hour')
```
pownkel, line 49: Can you also update the log line below to use `maxWallClockTimeInHours`?

pownkel (approved):
LGTM

PR #1318, opened 2021-03-05
fix: increase max memory allocated to chromium

pownkel:
#### Details

Increase the maximum threshold of memory available to Chromium to 8GB.

##### Motivation

For some URLs (ex. https://dotnet.microsoft.com/learn/dotnet/in-browser-tutorial/2), we see a sudden spike in memory use when scanning, which causes chromium to run out of memory and crash when running on the VMs. After this change, we can successfully scan those URLs.

#### Context

The current default memory limit for Chromium is dependent on the specific machine and environment. Our VMs have enough memory available to complete these scans, but it the current defaults prevent enough memory from being allocated to chromium.

#### Pull request checklist

### DIFF
--- a/packages/scanner-global-library/src/puppeteer-options.ts
+++ b/packages/scanner-global-library/src/puppeteer-options.ts
@@ -12,6 +12,6 @@ export const defaultBrowserOptions: Puppeteer.BrowserOptions = {
 
 export const defaultLaunchOptions: Puppeteer.LaunchOptions = {
     headless: true,
-    args: ['--disable-dev-shm-usage', '--no-sandbox', '--disable-setuid-sandbox'],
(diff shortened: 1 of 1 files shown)


packages/scanner-global-library/src/puppeteer-options.ts
```diff
@@ -12,6 +12,6 @@ export const defaultBrowserOptions: Puppeteer.BrowserOptions = {
 
 export const defaultLaunchOptions: Puppeteer.LaunchOptions = {
     headless: true,
-    args: ['--disable-dev-shm-usage', '--no-sandbox', '--disable-setuid-sandbox'],
+    args: ['--disable-dev-shm-usage', '--no-sandbox', '--disable-setuid-sandbox', '--js-flags=--max-old-space-size=8192'],
```
AhmedAbdoOrtiga, line 15: I think we should be fine with 4096, but it's ok

AhmedAbdoOrtiga (approved):

Issue #1295, opened 2021-02-17
[Bug] Investigate why some scans are not found 


AhmedAbdoOrtiga:
## Describe the bug

Some URLS gets deleted from either the batch scan container or the queue container before it make it to the permanent container.
That could be one of the following reasons:
  - The batch scan feed function failed for some reason (either cosmos trigger failed or internal error).
  - The url expired and deleted (7 days limit) in the queue container before it's picked up by the scan-request-sender.
  - Or it could be something else.

We need to come up with a design that retry to run batch-scan-feed function if it fails .

**Also we need to improve our telemetries as a part of solving this too:**
  - if the above scenario happens, we don't have enough telemetry to track what happened!

**This could end up a feature request**

## Area

- [ ] Install script
- [ ] Update script
- [x] Service

ferBonnin:
Per conversation with the team. to tackle with feature work. @ferBonnin to link to feature

PR #1319, opened 2021-03-08
Convert website scan result provider to support high concurrency write storage operation

### LINKED ISSUE #1306: [Bug] Resubmitted scan shows status stuck at In progress
## Describe the bug

Executed new scan to avoid the NotFound response issue in #1295 but the scan ids stuck with status In progress.

## Area

- [ ] Install script
- [ ] Update script
- [x] Service

## Expected behavior

Scan completes

lamaks:
#### Details

Converted the `WebsiteScanResult` document to be stored in multiple DB documents. The common data is stored in a single instance of the common `WebsiteScanResultBase` DB document with the composite key (`baseUrl`, `scanGroupId`). The frequent delta updates are stored in multiple `WebsiteScanResultPart` DB documents with the composite key (`baseId`, `scanId`). Each `WebsiteScanResultPart` DB document corresponds to the scan that created the update. The `WebsiteScanResult` document corresponding read operation will merge delta updates in a memory for consuming by application layer.

##### Motivation

Fix scan requests Cosmos DB trigger function processing error due to high concurrency write storage operation to the single `WebsiteScanResult` document instance

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #1306, #1295


packages/service-library/src/data-providers/website-scan-result-provider.ts
```diff
@@ -54,29 +69,43 @@ export class WebsiteScanResultProvider {
...
-                } else {
-                    return this.mergeAndWrite(operationResult.scanResult, websiteScanResult);
-                }
-            },
+            async () => this.mergeOrCreateImpl(dbDocument),
             async (err) =>
                 this.logger.logError(`Failed to update website scan result Cosmos DB document. Retrying on error.`, {
-                    document: JSON.stringify(scanResultNormalized),
+                    baseId: dbDocument.baseDocument.id,
+                    partId: dbDocument?.partDocument.id,
```
pownkel, line 80: Should the first line here also have a null check for dbDocument?

lamaks: updated

AhmedAbdoOrtiga (changes requested):
LGTM, can we submit many requests under same report group id? I just wanna simulate the case when the batch feed trigger fails because of concurrent t write to the same webRef document and make sure it's fixed

lamaks:
> LGTM, can we submit many requests under same report group id? I just wanna simulate the case when the batch feed trigger fails because of concurrent t write to the same webRef document and make sure it's fixed

Yes, this is what was tested and targeted for. The refactoring allow to skip writes for the same content db document and have frequent changes being written into separate db documents to avoid db document merge on write hence avoid write concurrency issue.

AhmedAbdoOrtiga (approved):

PR #1309, opened 2021-03-01
feat(host-browser-service): Configure Docker DNS

AhmedAbdoOrtiga:
#### Details
Configure docker DNS to resolve "host.docker.internal" to the host machine IP address.

#### Pull request checklist

- [ ] Addresses an existing issue: Fixes #0000
- [x] Added relevant unit test for your changes. (`yarn test`)
- [x] Verified code coverage for the changes made. Check coverage report at: `<rootDir>/test-results/unit/coverage`
- [x] Ran precheckin (`yarn precheckin`)
- [x] Validated in an Azure resource group

PR #1336, opened 2021-03-13
chore: increase VM reboot timeout and remove stale docker-dns script

karanbirsingh:
##### Motivation

- Increase timeout for batch nodes to become stable since the Windows infra can take longer. PPE and Prod environments have more nodes per pool so there's a higher chance of outliers. Follow-up from initial increase in PR #1310 
- Remove configure-docker-dns script added in #1309 This allowed the docker instance to communicate with the puppeteer browser server on the VM host. But since we switched back to scanning inside the Docker container, we don't need this script anymore.

#### Pull request checklist


pownkel (approved):
LGTM

PR #1373, opened 2021-04-09
feat(deep-scan-e2e-tests): add functional test groups for deep-scan

karanbirsingh:
#### Details

This PR adds some functional test groups for the deep-scan scenarios. These are not exercised yet, so this PR has no net behavior on the e2e test suite. We're just getting these files in ahead of time to simplify future reviews / testing.

##### Motivation

These functional test groups will be driven from the e2e-scan-scenario orchestrators in future PRs. This enables us to check whether each deep scan request:
- ends up crawling the expected set of URLs
- provides access to a report for each URL (consolidated will be checked using existing consolidated report test group)
- maintains an overall scan status that's consistent with individual URL scan states

##### Context

We thought about whether `expectedCrawledUrls?: string[];` should be provided other ways. Some alternatives include using snapshots, adding test group constructor args, etc. We decided to use the TestContextData for now to keep things simple. This is consistent with how the TestContextData looked before this feature - e.g. it had fields like consolidatedId, reportID that a subset of all functional tests required.

#### Pull request checklist


pownkel (approved):
LGTM

PR #1389, opened 2021-04-22
fix: update NavigationError processing to capture more specific errors

karanbirsingh:
#### Details

This PR updates our navigation-error processing so that more URLs fall into specific buckets like `UrlNavigationTimeout` or `SslError` rather than the catch-all `NavigationError`.

PR parts:
- small refactor of navigation-error mappings into separate file
- adding more error signatures like `net::ERR_CERT_AUTHORITY_INVALID` based on what we saw last WCP run
- update `Navigation timeout` string because of puppeteer update ([original code](https://github.com/puppeteer/puppeteer/blob/3773229ac276a84a4de113e74290abc3bbf60499/lib/LifecycleWatcher.js#L140), [new code](https://github.com/puppeteer/puppeteer/blob/18143b3573e3e9e637ae43c3f757bed8b752634f/src/common/LifecycleWatcher.ts#L194))
- Tested in Azure resource group with affected URLs from last month

##### Motivation

Specific failure cases help the user understand why we couldn't scan their website. Otherwise we need to check logs to get the root error.

#### Pull request checklist

### DIFF
--- a/packages/scanner-global-library/src/navigation-error-mappings.ts
+++ b/packages/scanner-global-library/src/navigation-error-mappings.ts
@@ -1,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { BrowserErrorTypes } from './browser-error';
+
+export const errorPatterns: Partial<Record<BrowserErrorTypes, string[]>> = {
+    UrlNavigationTimeout: ['Navigation timeout', 'net::ERR_CONNECTION_TIMED_OUT'],
+    SslError: [
+        'net::ERR_CERT_AUTHORITY_INVALID',
+        'net::ERR_CERT_DATE_INVALID',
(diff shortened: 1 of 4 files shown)


packages/scanner-global-library/src/navigation-error-mappings.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+import { BrowserErrorTypes } from './browser-error';
+
+export const errorPatterns: Partial<Record<BrowserErrorTypes, string[]>> = {
```
lamaks, line 5: inconsistent naming: navigation-error-mappings.ts , errorPatterns, navigationErrorPatterns (next file)

karanbirsingh: thanks, updated

lamaks (approved):

PR #1424, opened 2021-05-14
feat(axe-core-4.2.1): update axe-core and associated deps

dbjorge:
#### Details

This PR bumps `axe-core` and `accessibility-insights-report` to latest. It intentionally forces `@axe-core/puppeteer` to non-latest version 4.1.1 (the latest version does not support overriding the `allowedOrigins` axe-core configuration, and pinning to 4.1.1 works around this).

Besides the version bumps, the changes include:
* Fixing an issue in axe-puppeteer-factory which caused it to use a different version of axe-core than we pin to in cases where a consumer doesn't explicitly pass in a `contentSourcePath` (now, if that happens, we default to the source provided by us doing `import 'axe-core'` rather than allowing `@axe-core/puppeteer` to do that import, since it uses a less specific version range for `axe-core` than we do)
* Updating the rule exclusion list with the new needs-review/best-practice cases
* Setting up `scanner-global-library` to be able to invoke `.configure` while setting up `AxePuppeteer`
* Setting up separate default configurations for cloud/local cases (allowing all-origin access from the cloud workers and same-origin for local CLI/action cases)

##### Motivation

Keep web/service in sync

##### Context

We discussed the possibility of having the CLI/action accept an `allowedOrigins` parameter to override the same-origin default behavior; I've left this as out of scope for this PR. We can revisit that based on customer demand in CLI/action.

* [axe-core 4.2.0 release notes](https://github.com/dequelabs/axe-core/releases/tag/v4.2.0)
* [axe-core 4.2.1 release notes](https://github.com/dequelabs/axe-core/releases/tag/v4.2.1)
* [web 4.2.0 update PR](https://github.com/microsoft/accessibility-insights-web/pull/4155)
* [web 4.2.1 update PR](https://github.com/microsoft/accessibility-insights-web/pull/4241)

#### Pull request checklist

- [x] Addresses an existing issue: 1810733


dbjorge:
Ongoing manual validation in `danielbj-dev`:
* [x] E2E tests are passing
* [x] A shallow scan against [a test codepen](https://codepen.io/dbjorge/full/abJmWXM) with an axe-core 4.2.0-specific nested-interactive violation inside a cross-origin iframe is successfully identified as an error
* [ ] A deep scan against the same codepen also finds the same error inside the the cross-origin frame
* [ ] A build of the CLI package packed and referenced from a separate test project does *not* find that error (because it is in a cross-origin frame)
* [ ] A build of the CLI package packed and referenced from a separate test project *does* find that error for a test site with a nested-interactive violation inside a **same**-origin frame

dbjorge:
Ongoing manual validation in `danielbj-dev` against d42c457:
* [ ] E2E tests are passing
* [ ] A shallow scan against [a test codepen](https://codepen.io/dbjorge/full/abJmWXM) with an axe-core 4.2.0-specific nested-interactive violation inside a cross-origin iframe is successfully identified as an error
* [ ] A deep scan against the same codepen also finds the same error inside the the cross-origin frame
* [x] A build of the CLI package packed and referenced from a separate test project *does* find that error for a test site with a nested-interactive violation inside a **same**-origin frame
* [x] A build of the CLI package packed and referenced from a separate test project does *not* find that error (because it is in a cross-origin frame)

packages/scanner-global-library/src/setup-scanner-container.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function setupCloudScannerContainer(container: inversify.Container): inversify.Container {
+    container.bind(iocTypes.AxeConfiguration).toConstantValue(cloudAxeConfiguration);
+    container.bind(AxePuppeteerFactory).toSelf().inSingletonScope();
+
+    return container;
+}
+
+export function setupLocalScannerContainer(container: inversify.Container): inversify.Container {
+    container.bind(iocTypes.AxeConfiguration).toConstantValue(localAxeConfiguration);
```
pownkel, line 16: Do we need to bind AxePuppeteerFactory here as well?

pownkel (approved):
LGTM

packages/scanner-global-library/src/setup-scanner-container.ts
dbjorge: I don't think it makes any practical difference whether local uses a singleton or an instance-per-scan; I left that part of the logic as-is only because there didn't seem to be any compelling reason to change it.

PR #1474, opened 2021-06-08
Fix tslib version

lamaks:
#### Details

Fix tslib version to resolve runtime error: `tslib.__spreadArray is not a function`

[Deployment validation](https://portal.azure.com/?feature.customportal=false#@microsoft.onmicrosoft.com/resource/subscriptions/76eb140e-abd4-4a17-ad98-534d0f8d2759/resourceGroups/maxl-a11y-1/overview)

#### Pull request checklist


yarn.lock
```diff
@@ -12107,15 +12107,10 @@ tsconfig-paths@^3.9.0:
...
-tslib@^1.10.0, tslib@^1.13.0, tslib@^1.8.1, tslib@^1.9.0, tslib@^1.9.3:
-  version "1.13.0"
-  resolved "https://registry.yarnpkg.com/tslib/-/tslib-1.13.0.tgz#c881e13cc7015894ed914862d276436fa9a47043"
-  integrity sha512-i/6DQjL8Xf3be4K/E6Wgpekn5Qasl1usyw++dAA35Ue5orEn65VIxOA+YvNNl9HV3qv70T7CNwjODHZrLwvd1Q==
-
-tslib@^2.0.0:
-  version "2.0.1"
-  resolved "https://registry.yarnpkg.com/tslib/-/tslib-2.0.1.tgz#410eb0d113e5b6356490eec749603725b021b43e"
-  integrity sha512-SgIkNheinmEBgx1IUNirK0TUD4X9yjjBRTqqjggWCU3pUEqIk3/Uwl3yRixYKT6WjQuGiwDv4NomL3wqRCj+CQ==
+tslib@>=2.2.0, tslib@^1.10.0, tslib@^1.13.0, tslib@^1.8.1, tslib@^1.9.0, tslib@^1.9.3, tslib@^2.0.0:
```
dbjorge, line 12110: Would it be possible to use a more scoped fix than this? [The v2.0.0 release notes](https://github.com/microsoft/tslib/releases/tag/2.0.0) suggest that it includes a legitimate breaking change which is liable to subtly break packages that were built against typescript < 3.9. I feel nervous about blanket-updating so many transitive dependencies across that particular change boundary.

It would be helpful if you could include a little more context about the motivating error in the PR description; as-is, there isn't enough information to suggest a more scoped workaround.

lamaks: Downgrade to v1 to support TypeScript 3.8.4 or earlier

dbjorge: It looks like the updated resolution is for `"tslib": ">=1.14.1"`, but this doesn't actually downgrade to v1; that version spec will cause yarn to pick the greatest stable version above that point, so it's still going to pick up >v2 here. You can see that in the updated `yarn.lock` lines, where every tslib version in use is being resolved to v2.2.0:

```
tslib@>=1.14.1, tslib@^1.10.0, tslib@^1.13.0, tslib@^1.8.1, tslib@^1.9.0, tslib@^1.9.3, tslib@^2.0.0:
  version "2.2.0"
```

If updating the transitive dep within the bounds of what they claim to support is sufficient, you might not need a resolution here at all; deleting the resolution and tslib `yarn.lock` entry and then re-running `yarn install` would be enough.

If there really is a package that requires we force a major version bump, it would be preferable to limit the scope by using a more specific path on the left side of the resolution, like `"problematic/dependency/chain/tslib": "^2.2.0"` or `common-problematic-transitive-dependency/tslib": "^2.2.0"`

lamaks: Revert back as app insights update seems resolve issue

pownkel (approved):
LGTM

PR #1487, opened 2021-06-15
Patch security issues

lamaks:
#### Details

Patch security issues
Added yarn audit fix support

#### Pull request checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "scripts": {
         "bootstrap": "npx lerna bootstrap",
         "clean": "npx lerna run clean --stream && rimraf dist test-results",
-        "cbuild": "npx syncpack fix-mismatches format --indent '    ' && yarn install && npx lerna run build --stream",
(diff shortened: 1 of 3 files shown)


pownkel (dismissed):
LGTM

package.json
```diff
@@ -66,12 +68,9 @@
         "axe-core": "4.2.1",
         "apify-shared": ">=0.5.0",
         "axios": ">=0.21.1",
-        "socket.io": ">=2.4.0",
-        "socket.io-parser": ">=3.4.1",
```
dbjorge, line 70: The updated `yarn.lock` file now has dependencies on older versions of `socket.io-parser` and `engine.io` than the old resolutions allowed. Is that intended? (it looks like #1251 is the PR that originally added both of those, but unfortunately its description doesn't contain links to the specific CVEs it addressed and CG history only appears to go back 3 months)

dbjorge (changes requested):
Wanted to double check on two of the resolution removals

pownkel (approved):

dbjorge (approved):

PR #1496, opened 2021-06-16
Increase max number of failed scan requests retries

lamaks:
#### Details

Increase max number of failed scan retries

##### Motivation

Mitigate the concurrency issue when building consolidated report for websites with 100 pages 

##### Context

When website scan runs it queues all discovered pages to scan. Some pages will fail to update consolidated report blob at the same time. Increasing number of retries will span failed pages scan requests longer with more probability to update the report blob.

#### Pull request checklist


pownkel (approved):
LGTM

PR #1742, opened 2021-09-29
feat(baseline-implementation): Add singleWorker option to the scan engine

AhmedAbdoOrtiga:
#### Details

Allow the user to choose to run the scan engine using only a single crawler worker.

##### Motivation

This option will be used to help making the scan engine result deterministic, basically if the number of URLs in the website is > the MaxURLs, we should set singleWorker to true so we would get the same result every run.

##### Context
In the baselining feature, we need to make sure that the scan result is deterministic, so in the GH action and the ADO extension, we will set singleWorker to true.

#### Pull request checklist

### DIFF
--- a/packages/crawler/src/crawler/puppeteer-crawler-engine.ts
+++ b/packages/crawler/src/crawler/puppeteer-crawler-engine.ts
@@ -60,6 +60,11 @@ export class PuppeteerCrawlerEngine {
             this.crawlerConfiguration.setChromePath(crawlerRunOptions.chromePath);
         }
 
+        if (crawlerRunOptions.singleWorker === true) {
+            puppeteerCrawlerOptions.minConcurrency = 1;
+            puppeteerCrawlerOptions.maxConcurrency = 1;
+        }
+
         if (crawlerRunOptions.debug === true) {
             this.crawlerConfiguration.setSilentMode(false);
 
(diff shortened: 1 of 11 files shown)


packages/crawler/src/crawler/puppeteer-crawler-engine.ts
```diff
@@ -60,6 +60,11 @@ export class PuppeteerCrawlerEngine {
             this.crawlerConfiguration.setChromePath(crawlerRunOptions.chromePath);
         }
 
+        if (crawlerRunOptions.singleWorker === true) {
```
DaveTryon, line 63: Do we have unit test code that validates this behavior? The current test code replicates this behavior but that's not the same as validating it.

AhmedAbdoOrtiga: yes we do, in the PR I added new test cases to validate this code

packages/crawler/src/crawler/simple-crawler-engine.ts
```diff
@@ -37,6 +37,11 @@ export class SimpleCrawlerEngine implements CrawlerEngine<string[]> {
             maxRequestsPerCrawl: this.crawlerConfiguration.maxRequestsPerCrawl(),
         };
 
+        if (crawlerRunOptions.singleWorker === true) {
```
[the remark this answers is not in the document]
AhmedAbdoOrtiga: yes we do, in the PR I added new test cases to validate this code

DaveTryon (approved):

PR #2000, opened 2022-03-17
chore: move syncpack checks from cbuild to more specific run scripts

dbjorge:
#### Details

Before this PR, the root level `package.json`'s `cbuild` run script incorporated syncpack like this:

```
"cbuild": "npx syncpack fix-mismatches format --indent '    ' && yarn install && npx lerna run cbuild --stream",
```

This is problematic for several different reasons (see Motivation section). This PR moves the syncpack commands out of `cbuild`, and instead puts the "check for mismatches" part in `precheckin` and a separate CI build step, the "fix formatting" part into "format:fix", and the "fix mismatches" part into a separate explicit command. It also fixes the syntax of the `--indent` argument in the "format" part to be compatible with Windows/PowerShell environments.

##### Motivation

1. The `syncpack fix-mismatches` command is something we really wouldn't want a CI/CD build to ever run accidentally; it would run the risk of the CI/CD build producing output which mismatched the versions defined in source. However, `cbuild` sounds like a command that would be reasonable to run from a workflow/pipeline. Moving the command that can mutate source files into an explicit `fix` run script makes mistakes less likely
2. Same goes for the `yarn install` command without `--frozen-lockfile` - we want it to be very hard to accidentally run a non-frozen yarn install in CI builds
3. Previously, we weren't actually enforcing the version syncing in CI builds, only in local developer cbuilds. Now we enforce them both in CI builds and in `yarn precheckin`.
4. This incidentally fixes an incompatibility with running the `cbuild` command from Windows machines - the previous syntax used `--indent '    '`, which does the wrong thing in a Windows/PowerShell environment (it uses `'    '` instead of `    ` to indent). Besides this issue (which is fixed by instead using `--indent \"    \"`, syncpack also blindly assumes that you want platform-native EOLs, which isn't necessarily true on Windows depending on your git config; bundling it into `format:fix` with the prettier command second ensures that prettier will fix newlines appropriately even if syncpack gets them wrong.

##### Context

n/a

#### Pull request checklist


dbjorge:
The PR build failure is catching a legitimate issue with mismatched versions introduced with #1989. Submitted a separate PR #2001 to address this; will rebase this one on top of #2001 after it merges.

package.json
```diff
@@ -19,13 +19,15 @@
         "test:e2e": "npx jest functional-tests.local.spec.ts",
         "lint:check": "npx lerna run lint --stream",
         "lint:fix": "npx lerna run lint:fix --stream",
-        "precheckin": "npm-run-all --serial format:check lint:check copyright:check cbuild test",
-        "format:fix": "npx prettier --write \"**/*\"",
+        "precheckin": "npm-run-all --serial format:check lint:check copyright:check syncpack:check cbuild test",
+        "format:fix": "npx syncpack format --indent \"    \" && npx prettier --write \"**/*\"",
```
lamaks, line 23: can we add [configuration file](https://github.com/JamieMason/syncpack#-configuration-file) and remove --indent as well?

dbjorge: On further testing, I decided to just remove the `syncpack format` call entirely because

1. syncpack doesn't currently offer any `format:check` equivalent to enforce it in CI builds
2. none of our current `package.json` files actually respect its formatting today (it looks like the previous `cbuild`-based command didn't actually perform any of the non-indent reformatting when `fix-mismatches` and `format` were in the same command line)

We could implement our own script to do a format:check based on whether format + prettier changes anything, but if we wanted to go that route, I think it would be preferable to instead contribute upstream in https://github.com/JamieMason/syncpack/issues/3

lamaks: ok to drop, we will rely on current formatting mechanism

lamaks (approved):

package.json
JamieMason: Hi all, as of [12.3.0](https://github.com/JamieMason/syncpack/releases/tag/12.3.0) `syncpack lint` will now also lint formatting.

PR #2011, opened 2022-03-28
(Node 16 update) Apify 2.2.2, Puppeteer 10.0.0, and Node 16

pownkel:
#### Details

Significant updates:
- Migrate from Apify 0.21.8 to 2.2.2 (Involves significant changes to the Apify interface)
- Migrate from Puppeteer 5.4.0 to 10.0.0
- Migrate from Node 12 to Node 16

Other changes:
- Bump the CLI package version from 0.10.4 to 0.11.0 (breaking change: package now requires Node 16)
- With Puppeteer 10, some pages do not load all offscreen content, but scanning doesn't fail like we expect it to in this case. To solve this, we scroll to the bottom of the page each time before scanning.

##### Motivation

Node 12 will soon reach EOL. Our current version of Apify constrains us to Node 12. Updating to the most recent Apify package (2.2.2) also requires updates to Node >=16 and puppeteer, so all three updates must be made simultaneously.

##### Context

I also explored migrating to Apify v1 and Node 14, which would fix the issue of moving us off a soon-to-be-EOL version of node. However, Apify v1 has some [unexpected behavior](https://github.com/apify/apify-js/issues/1002) when configuring the crawler through environment variables, which broke our implementation for setting a custom output directory. Apify 1.3.4 introduced a way to configure the SDK without using environment variables, but using a configured SDK caused [SQLite errors](https://github.com/apify/apify-js/issues/1305).

The old env variable behavior was restored in Apify 2, so we currently still use env variables to configure Apify. This method is a bit clunky, though, so we may want to use a configured SDK instead after the fix for the above linked bug is released.

#### Pull request checklist


packages/cli/package.json
```diff
@@ -81,7 +81,8 @@
         "lodash": "^4.17.21",
         "moment": "^2.29.1",
         "normalize-path": "^3.0.0",
-        "puppeteer": "^5.5.0",
+        "playwright": "^1.19.2",
```
lamaks, line 84: where do we use playwright?

packages/scanner-global-library/src/navigation-hooks.ts
```diff
@@ -15,7 +15,10 @@ export class NavigationHooks {
         @inject(PageConfigurator) public readonly pageConfigurator: PageConfigurator,
         @inject(PageResponseProcessor) protected readonly pageResponseProcessor: PageResponseProcessor,
         @inject(PageHandler) protected readonly pageRenderingHandler: PageHandler,
+        private readonly scrollTimeoutMsecs = 15000,
```
lamaks, line 18: we can increase both `scrollTimeoutMsecs` and `pageRenderingTimeoutMsecs` to 30 secs and make them equal

packages/scanner-global-library/src/page-handler.ts
```diff
@@ -8,14 +8,50 @@ import { System } from 'common';
...
+        await this.scrollToBottom(page, scrollTimeoutMsecs);
+        await this.waitForStableContent(page, renderTimeoutMsecs);
+    }
+
+    private async scrollToBottom(page: Puppeteer.Page, timeoutMsecs: number): Promise<void> {
+        const maxCheckCount = timeoutMsecs / this.checkIntervalMsecs;
+        let checkCount = 0;
+        let scrollingComplete = false;
+
+        // Scroll incrementally so everything is inside the window at at some point
```
lamaks, line 31:
```suggestion
        // Scroll incrementally so everything is inside the window at some point
```

packages/scanner-global-library/src/page-handler.ts
```diff
@@ -8,14 +8,50 @@ import { System } from 'common';
...
+        let scrollingComplete = false;
+
+        // Scroll incrementally so everything is inside the window at some point
+        while (!scrollingComplete && checkCount < maxCheckCount) {
+            // Page evaluation may fail because of a navigation
+            try {
+                scrollingComplete = await page.evaluate(async () => {
+                    window.scrollBy(0, window.innerHeight);
+
+                    return window.document.scrollingElement.scrollTop + window.innerHeight >= window.document.scrollingElement.scrollHeight;
```
lamaks, line 38: do we need `return`?

packages/web-api-scan-runner/package.json
```diff
@@ -69,13 +67,14 @@
...
         "p-limit": "^3.1.0",
-        "puppeteer": "^5.5.0",
+        "puppeteer": "^10.0.0",
         "reflect-metadata": "^0.1.13",
         "scanner-global-library": "1.0.0",
         "service-library": "1.0.0",
         "sha.js": "^2.4.11",
         "storage-documents": "1.0.0",
         "verror": "^1.10.1",
+        "yamlparser": "^0.0.2",
```
lamaks, line 77: where do we use `yamlparser`?

packages/cli/package.json
pownkel: It's an optional dependency from Apify but it's referenced in Apify's typings, so the build breaks if we don't install it ourselves

packages/scanner-global-library/src/page-handler.ts
pownkel: Yes, this is returning a boolean value for whether or not we've reached the end of the page

packages/web-api-scan-runner/package.json
pownkel: Similar to playwright, this is an incorrectly declared dependency in a transitive dependency that is needed for building with webpack to work

lamaks (approved):

PR #2056, opened 2022-04-14
Fixed HTTP response size normalization

lamaks:
#### Details

Fixed HTTP response normalization to reduce log size
Fixed stale running report requests removal from a queue
Added logging to simplify debugging

#### Pull request checklist

### DIFF
--- a/packages/common/src/system/system.spec.ts
+++ b/packages/common/src/system/system.spec.ts
@@ -107,25 +107,31 @@ describe('serializeError()', () => {
 });
 
 describe('normalizeHttpResponse()', () => {
-    it('should normalize request object', () => {
+    it('should normalize HTTP request object', () => {
+        const requestObj = {
+            url: 'url',
+            method: 'PUT',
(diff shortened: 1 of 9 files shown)


packages/common/src/system/system.spec.ts
```diff
@@ -107,25 +107,31 @@ describe('serializeError()', () => {
...
+                status: 412,
+                bodyAsText: 'bodyAsText',
             },
         } as any;
-        const { request, ...httpResponseExpected } = httpResponse;
-        httpResponseExpected.request = {
-            url: 'request url',
-            method: 'PUT',
-        };
+        const expectedResponse = JSON.parse(
```
pownkel, line 130: Why is JSON.parse called here instead of creating the object?

lamaks: to simplify test code, similar to snapshot

pownkel (approved):

PR #2059, opened 2022-04-22
Fixed page navigation logic

lamaks:
#### Details

Updated page navigation logic to reduce navigation wait time and false failure path
Updated page scrolling logic to ensure scroll completion detection
Fixed page resizing to match browser viewport

#### Pull request checklist

### DIFF
--- a/packages/scanner-global-library/src/page-configurator.ts
+++ b/packages/scanner-global-library/src/page-configurator.ts
@@ -3,33 +3,28 @@
 
 import { injectable } from 'inversify';
 import * as Puppeteer from 'puppeteer';
+import { defaultBrowserOptions } from './puppeteer-options';
 
 @injectable()
 export class PageConfigurator {
     private userAgent: string;
-
-    private readonly windowWidth = 1920;
-
-    private readonly windowHeight = 1080;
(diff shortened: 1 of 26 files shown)


packages/scanner-global-library/src/page-configurator.ts
```diff
@@ -3,33 +3,28 @@
...
-    private readonly windowHeight = 1080;
-
     public getUserAgent(): string {
         return this.userAgent;
     }
 
     public async configurePage(page: Puppeteer.Page): Promise<void> {
+        await page.setViewport(defaultBrowserOptions.defaultViewport);
+        //@ts-expect-error
+        await page._client.send('Emulation.clearDeviceMetricsOverride'); // enable page resizing to match to browser viewport
```
pownkel, line 19: Can `page._client` this be replaced with `page.target().createCDPSession()` so we aren't relying on private fields?

lamaks: No, session did not work, no other solution but using `_client` property. It is part of page.d.ts definition.

pownkel (approved):

PR #2093, opened 2022-05-17
feat(ado-auth-part-1):  Enable crawler to scan authenticated pages using MSFT SSO authentication

madalynrose:
#### Details

* allows users to specify service account name and password in crawler parameters `--serviceAccountName` and `--serviceAccountPassword`
* implements browser authentication flow for `portal.azure.com` to be executed after a browser session launches (just before crawling begins)
* surfaces authentication errors to the user via logs
* adds unit tests for authentication steps

##### Motivation

Feature work

##### Context

This functionality is intended to be consumed by the `accessibility-insights-action` ADO extension to allow scanning of authenticated pages.

We are using `portal.azure.com` as a reliable starting point for Microsoft single sign-on authentication using a service account's email and password. This authentication flow requires a connection to a known Microsoft network.

#### Pull request checklist

- [ n/a] Addresses an existing issue: Fixes #0000


packages/cli/src/crawler/crawler-parameters-builder.ts
```diff
@@ -43,6 +43,8 @@ export class CrawlerParametersBuilder {
             axeSourcePath: scanArguments.axeSourcePath,
             debug: scanArguments.debug,
             singleWorker: scanArguments.singleWorker,
+            serviceAccountName: scanArguments.serviceAccountName,
+            serviceAccountPass: scanArguments.serviceAccountPass,
```
lamaks, line 47: prefer full name over abbreviation `serviceAccountPassword`

packages/cli/src/scan-arguments.ts
```diff
@@ -23,4 +23,6 @@ export interface ScanArguments {
     axeSourcePath?: string;
     debug?: boolean;
     singleWorker?: boolean;
+    serviceAccountName?: string;
+    serviceAccountPass?: string;
```
lamaks, line 27: prefer full name over abbreviation `serviceAccountPassword`

packages/cli/src/validate-scan-arguments.ts
```diff
@@ -24,4 +24,15 @@ export function validateScanArguments(args: ScanArguments): void {
     if (args.updateBaseline === true && isEmpty(args.baselineFile)) {
         throw new Error('Option --updateBaseline requires option --baselineFile.');
     }
+
+    if (!args.crawl && (!isEmpty(args.serviceAccountName) || !isEmpty(args.serviceAccountPass))) {
+        throw new Error('Options --serviceAccountName and --serviceAccountPass are only supported with --crawl.');
```
lamaks, line 29: do you mean `supported with --crawl option`?
update to `serviceAccountPassword`

packages/cli/src/validate-scan-arguments.ts
```diff
@@ -24,4 +24,15 @@ export function validateScanArguments(args: ScanArguments): void {
...
+    }
+
+    if (
+        (isEmpty(args.serviceAccountName) && !isEmpty(args.serviceAccountPass)) ||
+        (!isEmpty(args.serviceAccountName) && isEmpty(args.serviceAccountPass))
+    ) {
+        throw new Error('Both --serviceAccountName and --serviceAccountPass must be provided for authenticated crawling.');
```
lamaks, line 36: do you mean `must be provided for website authentication`?

packages/crawler/src/authenticator/browser-authenticator.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+            .verifiable(Times.exactly(1));
+    });
+
+    it('follows portal.azure.com authentication flow', async () => {
+        setupPortalAuthenticationFlow(pageMock, keyboardMock, accountName, accountPass, loggerMock);
+        await authenticateBrowser(browserMock.object, accountName, accountPass, loggerMock.object);
+        browserMock.verifyAll();
+        pageMock.verifyAll();
+        keyboardMock.verifyAll();
+        loggerMock.verifyAll();
```
lamaks, line 67: can be moved into `afterEach()`

packages/crawler/src/authenticator/browser-authenticator.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    });
+
+    it('retries four times if it detects authentication failed', async () => {
+        setupPortalAuthenticationFlow(pageMock, keyboardMock, accountName, accountPass, loggerMock, false, 5);
+        await authenticateBrowser(browserMock.object, accountName, accountPass, loggerMock.object);
+        browserMock.verifyAll();
+        pageMock.verifyAll();
+        keyboardMock.verifyAll();
+        loggerMock.verifyAll();
```
lamaks, line 76: can be moved into `afterEach()`

packages/crawler/src/authenticator/browser-authenticator.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+
+import { IMock, It, Mock, Times } from 'typemoq';
+import { getPromisableDynamicMock } from '../test-utilities/promisable-mock';
+import { authenticateBrowser } from './browser-authenticator';
+
+const setupPortalAuthenticationFlow = (
+    pageMock: IMock<Puppeteer.Page>,
+    keyboardMock: IMock<Puppeteer.Keyboard>,
+    accountName: string,
+    accountPass: string,
```
lamaks, line 16: prefer full name over abbreviation `accountPassword`

packages/crawler/src/authenticator/browser-authenticator.spec.ts
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import Apify from 'apify';
+const { log: apifyLog } = Apify.utils;
```
lamaks, line 5: should not be part of imports block

packages/crawler/src/authenticator/browser-authenticator.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+import Puppeteer from 'puppeteer';
+
```
lamaks, line 7: neat: empty line

packages/crawler/src/authenticator/browser-authenticator.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+import { IMock, It, Mock, Times } from 'typemoq';
+import { getPromisableDynamicMock } from '../test-utilities/promisable-mock';
+import { authenticateBrowser } from './browser-authenticator';
+
+const setupPortalAuthenticationFlow = (
```
lamaks, line 12: why we need it as a `const` instead of function?

packages/crawler/src/authenticator/browser-authenticator.ts
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+import Apify from 'apify';
+const { log: apifyLog } = Apify.utils;
```
lamaks, line 5: `const` not part of imports block

packages/crawler/src/authenticator/browser-authenticator.ts
```diff
@@ -0,0 +1,51 @@
...
+import * as Puppeteer from 'puppeteer';
+
+const attemptAuthentication = async (
+    page: Puppeteer.Page,
+    accountName: string,
+    accountPass: string,
+    logger: typeof apifyLog,
+    attemptNumber: number = 1,
+): Promise<void> => {
+    await page.goto('https://portal.azure.com');
```
lamaks, line 15: why this URL is hardcoded here?

packages/crawler/src/types/crawler-run-options.ts
```diff
@@ -21,4 +21,6 @@ export interface CrawlerRunOptions {
     chromePath?: string;
     axeSourcePath?: string;
     singleWorker?: boolean;
+    serviceAccountName?: string;
+    serviceAccountPass?: string;
```
lamaks, line 25: `serviceAccountPassword`?

packages/crawler/src/authenticator/browser-authenticator.ts
```diff
@@ -0,0 +1,51 @@
...
+    await page.waitForSelector('#FormsAuthentication');
+    await page.click('#FormsAuthentication');
+    await page.type('input[type="password"]', accountPass);
+
+    await page.keyboard.press('Enter');
+    await page.waitForNavigation({ waitUntil: 'networkidle0' });
+    if (!page.url().match('^https://ms.portal.azure.com')) {
+        const errorText: string = await page.$eval('#errorText', (el) => el.textContent);
+        if (attemptNumber > 4) {
+            logger.error('Attempted authentication 5 times and ultimately failed.');
```
lamaks, line 28: hardcoded value should not be used `5 times`

packages/crawler/src/authenticator/browser-authenticator.ts
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
lamaks, line 1: This is specific implementation for the portal.azure.com. The better approach would be to have generic interface and specific implementation with corresponding class name. May be create auth provider based on auth website underlying mechanism?

Defining a function and function signature type instead of `const` will help. This code is more javascript rather than typescript.

madalynrose: > This is specific implementation for the portal.azure.com. The better approach would be to have generic interface and specific implementation with corresponding class name.

Absolutely agree! However, we don't plan to support other authentication flows at this time. It seemed overengineered to add an unnecessary abstraction layer when our goal is to always use Microsoft SSO inside a Microsoft network, limited to usage with our ADO Extension.

packages/cli/src/validate-scan-arguments.ts
madalynrose: I was matching this existing error message: 
```
    if (!args.crawl && !isEmpty(args.baselineFile)) {
        throw new Error('Option --baselineFile is only supported with --crawl.');
    }
```

Should I change both or leave it?

packages/cli/src/validate-scan-arguments.ts
madalynrose: what about `must be provided to scan authenticated pages` ?

packages/crawler/src/authenticator/browser-authenticator.ts
brocktaylor7: Hey Madalyn! Sorry, you may not have had the full context here, it was likely never discussed in any meetings with the whole team. 

We do want to add at least dSTS auth in the future (That's what is used when logging in with a PME account), and also possibly the login.live.com auth, which is outside of internal AAD and wouldn't work with the AAD SSO we have implemented here. So it's probably justifiable to create an abstraction here.

packages/cli/src/validate-scan-arguments.ts
lamaks: ok

packages/cli/src/validate-scan-arguments.ts
lamaks: we can update baselineFile message as well to match

packages/crawler/src/authenticator/browser-authenticator.spec.ts
lamaks: If mock object is not setup for the particular test `verifyAll()` will not break the test, generally setting mock with `verifiable()` will define validation logic.

packages/crawler/src/authenticator/browser-authenticator.ts
lamaks: It's okay to keep it inline if we have concrete implementation for azure provider only under some provider interface generalization.

packages/crawler/src/authenticator/authenticator.ts
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        if (!this.checkAuthenticationSucceeded(page, this.authenticationFlow.authenticatedUrl)) {
+            const errorText: string = await page.$eval('#errorText', (el) => el.textContent);
+            if (attemptNumber > 4) {
+                this.logger.error(`Attempted authentication ${attemptNumber} times and ultimately failed.`);
+
+                return;
+            }
+            if (errorText !== '') {
```
lamaks, line 54: will be true for `errorText === undefined`. IsEmpty (from lodash) will work better.

packages/crawler/src/authenticator/authenticator.ts
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under the MIT License.
+
+import Apify from 'apify';
+import * as Puppeteer from 'puppeteer';
+import { AuthenticationStep, AuthenticationFlow } from './authentication-flow';
+
+export class Authenticator {
+    public constructor(
+        protected authenticationFlow: AuthenticationFlow,
+        private readonly logger: typeof Apify.utils.log = Apify.utils.log,
```
lamaks, line 11: we can use console.log() as we do across crawler package. Do we have any benefits from Apify logger?

packages/crawler/src/authenticator/authenticator.ts
```diff
@@ -0,0 +1,67 @@
...
+    ) {}
+
+    public async run(browser: Puppeteer.Browser): Promise<void> {
+        const page = await browser.newPage();
+        await this.attemptAuthentication(page);
+        await page.close();
+    }
+
+    private async executeAuthenticationStep(page: Puppeteer.Page, step: AuthenticationStep): Promise<void> {
```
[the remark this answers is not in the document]
lamaks: We do have dependencies on puppeteer across entire solution as well as axe core depends on puppeteer. We can do more abstraction over auth class like adding some interface, but probably will need have a concrete implementation using puppeteer page.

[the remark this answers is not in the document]
madalynrose: Good points @brocktaylor7! I was aiming to consolidate as much hard-coded flow information as possible and keep actual functionality pure and testable. To make the abstraction less rigid, I'll opt for the interface with an `attemptAuthentication` method and put all of the hardcoded steps there.

packages/crawler/src/authenticator/authenticator.ts
madalynrose: this was so we could mock and test our log output.  If this is unnecessary I can use console.log

lamaks: console.log() probably will have less dependency footprint.

brocktaylor7 (approved):
LGTM, this is great work, Madalyn!

PR #2270, opened 2022-10-26
Update authentication to throw errors and catch account name errors

katydecorah:
#### Details

This pull request will make improvements to authentication errors:

- throw errors in the Azure active directory flow to stop the scanner from continuing. Previously, if unable to authenticate, the scanner would log the error and continue to scan the login page; resulting in a false positive.
- add a try/catch statement when entering the account name and throw a specific error from the login page. Previously, if a pipeline entered an invalid account name, it would return a timeout error.
- remove attempts. Now that auth will throw errors, the `handleRequestFunction` will retry up to 3 times and removes the need for auth to kick off additional attempts.

##### Motivation

Improve user experience when authenticating.

#### Pull request checklist


ThanyaLeif (approved):
Tested locally and I got the "Authentication failed with error:..." message error.
Well done!!

PR #2295, opened 2022-11-09
fix: Update auth flow to handle KMSI page

brocktaylor7:
#### Details

This adds a check for the KMSI (Keep Me Signed In) page that shows up conditionally based on a machine learning algorithm that detects high risk logins. It also bumps the scan package version in preparation for release.

More information on the KMSI page can be found here: https://learn.microsoft.com/en-us/azure/active-directory/fundamentals/customize-branding#learn-about-the-stay-signed-in-prompt

##### Motivation

Fix for auth

##### Context

My initial implementation did not use the try/catch block, but instead was performing a `page.$eval` and looking for the selector. I changed it to a `waitForSelector` with a 5000ms timeout to ensure that we aren't missing the dialog due to a possible race condition where the page doesn't fully render before the `page.$eval` executes.

#### Pull request checklist

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
     "name": "accessibility-insights-scan",
-    "version": "1.5.1",
+    "version": "1.5.2",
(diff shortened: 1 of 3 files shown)


ThanyaLeif (approved):
LGTM! 🚢

PR #2296, opened 2022-11-14
Ensure that the most precise error information is included in a scan run result

alisonlomaka:
#### Details
A previous change serialized a BrowserError to string before passing it into Runner.setRunResult. This had the unintended side effect of resulting in all BrowserErrors being reported as InternalError, instead of being mapped to the corresponding ScanRunErrorCode.

This change explicitly converts Error and BrowserError objects into ScanError objects before passing them to setRunResult. This works because both BrowserError and ScanError objects have an errorType member and the BrowserErrorTypes and ScanErrorTypes align. Error has no overlap with ScanError so it is mapped to InternalError - which is what the conversion handling would ultimately do with a string anyway.

For reference see also ScanResponseConverter.createFailedScanResponse and ScanErrorConverter.getScanRunErrorCode; this is the code path whereby a string | ScanError is converted into a ScanRunError for inclusion in a ScanResultResponse.

##### Motivation
Addresses internal bug 2007185 written against WCP.

##### Context
This was working before by a lucky loophole. We were passing a variable of type string | BrowserError into a type of string | ScanError. The string types on both sides allowed it to pass compile but then sneak through a BrowserError at runtime, which happened to be assignable and consumable as a ScanError.

I don't love this solution but it does make the luck and assumptions explicit, and it is narrowly scoped. Any change to the actual types defined for OnDemandPageScanResult or ScanResultResponse would have had a much wider ripple effect including the potential to affect a11y as well.

#### Pull request checklist

- [x] Addresses an existing issue: Fixes #0000

### DIFF
--- a/packages/privacy-scan-runner/src/runner/runner.spec.ts
+++ b/packages/privacy-scan-runner/src/runner/runner.spec.ts
@@ -246,10 +246,10 @@ function setupProcessScanResult(): void {
         if ((privacyScanResults.error as BrowserError)?.errorType === 'BannerXPathNotDetected') {
             runState = pageScanResult.run?.retryCount >= maxFailedScanRetryCount ? 'completed' : 'failed';
         }
-        const scanError = privacyScanResults.error as ScanError;
-        if (isEmpty(scanError.errorType)) {
-            scanError.errorType = 'InternalError';
-        }
+        const scanError: ScanError = {
+            errorType: 'InternalError'
+            ...privacyScanResults.error,
+        };
         pageScanResult.run = {
             state: runState,
             timestamp: dateNow.toJSON(),
--- a/packages/privacy-scan-runner/src/runner/runner.spec.ts
+++ b/packages/privacy-scan-runner/src/runner/runner.spec.ts
@@ -13,7 +13,6 @@ import {
 } from 'service-library';
 import { GlobalLogger } from 'logger';
 import * as MockDate from 'mockdate';
-import { isEmpty } from 'lodash';
 import {
     OnDemandPageScanResult,
     OnDemandPageScanReport,
@@ -247,7 +246,7 @@ function setupProcessScanResult(): void {
             runState = pageScanResult.run?.retryCount >= maxFailedScanRetryCount ? 'completed' : 'failed';
         }
         const scanError: ScanError = {
-            errorType: 'InternalError'
+            errorType: 'InternalError',
             ...privacyScanResults.error,
         };
         pageScanResult.run = {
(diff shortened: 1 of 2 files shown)


packages/privacy-scan-runner/src/runner/runner.spec.ts
```diff
@@ -244,10 +246,14 @@ function setupProcessScanResult(): void {
         if ((privacyScanResults.error as BrowserError)?.errorType === 'BannerXPathNotDetected') {
             runState = pageScanResult.run?.retryCount >= maxFailedScanRetryCount ? 'completed' : 'failed';
         }
+        const scanError = privacyScanResults.error as ScanError;
+        if (isEmpty(scanError.errorType)) {
+            scanError.errorType = 'InternalError';
+        }
```
pownkel, line 252: Nit (errorType will be overwritten if it exists in privacyScanResults.error)
```suggestion
        const scanError: ScanError = {
            errorType: 'InternalError'
            ...privacyScanResults.error,
        };
```

packages/privacy-scan-runner/src/runner/runner.spec.ts
```diff
@@ -244,10 +246,14 @@ function setupProcessScanResult(): void {
         if ((privacyScanResults.error as BrowserError)?.errorType === 'BannerXPathNotDetected') {
             runState = pageScanResult.run?.retryCount >= maxFailedScanRetryCount ? 'completed' : 'failed';
         }
+        const scanError = privacyScanResults.error as ScanError;
+        if (isEmpty(scanError.errorType)) {
+            scanError.errorType = 'InternalError';
```
pownkel, line 251: Is it still possible for privacyScanResults.error to be a string? What happens here in that case?

packages/privacy-scan-runner/src/runner/runner.ts
```diff
@@ -186,6 +186,15 @@ export class Runner {
         return !isEmpty(reports) ? this.reportWriter.writeBatch(reports) : undefined;
     }
 
+    private convertToScanError(error: Error | BrowserError): ScanError {
+        const scanError = error as ScanError;
+        if (isEmpty(scanError.errorType)) {
+            scanError.errorType = 'InternalError';
+        }
```
pownkel, line 193: Nit
```suggestion
        const scanError: ScanError = {
            errorType: 'InternalError',
            ...error,
        };
```

packages/privacy-scan-runner/src/runner/runner.spec.ts
alisonlomaka: It is not; Maxim's previous changes modified the signature for PrivacyScanResult.error to be Error | BrowserError.

pownkel (approved):

PR #2363, opened 2023-01-27
Add secret names for scan-behind-auth service account credentials.

alisonlomaka:
#### Details

Add values to secretNames for scanBehindAuthUsername and scanBehindAuthPassword. These correspond to the names of secrets in the service key vault. The secret value for scanBehindAuthUsername is expected to be the alias of a valid account to be used when scanning sites that require authentication. The secret value for scanBehindAuthPassword is the corresponding password for the account.

The secrets are configured manually on the key vault at or after creation time; since the key vault persists through redeployments this is not expected to add a maintenance burden.

##### Motivation

Allow the service to securely retrieve account credentials for authenticating sites at runtime.

#### Pull request checklist

### DIFF
--- a/packages/azure-services/src/key-vault/secret-names.ts
+++ b/packages/azure-services/src/key-vault/secret-names.ts
@@ -9,4 +9,6 @@ export const secretNames = {
     restApiSpSecret: 'restApiSpSecret',
     authorityUrl: 'authorityUrl',
     appInsightsApiKey: 'appInsightsApiKey',
+    scanBehindAuthUsername: 'scanBehindAuthUsername',
+    scanBehindAuthPassword: 'scanBehindAuthPassword',
 };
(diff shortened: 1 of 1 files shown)


packages/azure-services/src/key-vault/secret-names.ts
```diff
@@ -9,4 +9,6 @@ export const secretNames = {
     restApiSpSecret: 'restApiSpSecret',
     authorityUrl: 'authorityUrl',
     appInsightsApiKey: 'appInsightsApiKey',
+    scanBehindAuthUsername: 'scanBehindAuthUsername',
```
lamaks, line 12:
```suggestion
    azureServicePrincipalAuthClientName: 'scanBehindAuthUsername',
```

packages/azure-services/src/key-vault/secret-names.ts
```diff
@@ -9,4 +9,6 @@ export const secretNames = {
...
+    scanBehindAuthPassword: 'scanBehindAuthPassword',
```
lamaks, line 13:
```suggestion
    azureServicePrincipalAuthClientPassword: 'scanBehindAuthPassword',
```

lamaks (approved):

PR #2489, opened 2023-10-04
Allow processing authenticated URL

lamaks:
#### Details

Allow processing authenticated URL to enable scan with authentication.

#### Pull request checklist


adsathye (approved):

packages/api-contracts/ai-scan-api.yaml
```diff
@@ -349,17 +359,18 @@ components:
                 code:
                     type: string
                     enum:
-                        - InternalError
```
alisonlomaka, line 352: Is InternalError actually going away? I did not see which change in the PR would do this but it would be a breaking change for clients.

lamaks: It is service OpenAPI specification file I've updated. Just added missing enum values to have parity with code.

alisonlomaka (approved):

PR #2524, opened 2024-02-09
Enable crawling for URLs having hash fragments for ai-scan cli

v-viyada:
#### Details

Added new optional parameter keepUrlFragment to enable crawling for URLs having hash fragments for ai-scan cli and crawler packages.
The default value of it is false so that existing scan work as is.

##### Motivation

[Feature 2114928](https://dev.azure.com/mseng/1ES/_workitems/edit/2114928): ADO Extension should be able to crawl hash-routing SPA apps

#### Pull request checklist

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -290,3 +290,10 @@ The URL folder is a resource location equal to base URL up-to the last forward s
 type: array
 describe: List of Chrome command line options to pass on browser start. Can be used to disable CORS to scan protected page: --browserOptions disable-web-security
 ```
+
+-   keepUrlFragment: --keepUrlFragment
+
+```sh
+type: boolean
+describe: To keep the hash fragment in the URLs. If set to false, it will remove the hash fragment from URL. For example, http://www.example.com/#foo will be considered as http://www.example.com. Default value is false.
+```
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -85,6 +85,13 @@ describe: Output directory. If not set, default is ./ai_scan_cli_output, if you 
 default: './ai_scan_cli_output'
 ```
 
+-   keepUrlFragment: --keepUrlFragment
+
+```sh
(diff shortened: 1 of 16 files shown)


packages/cli/README.md
```diff
@@ -290,3 +290,10 @@ The URL folder is a resource location equal to base URL up-to the last forward s
 type: array
 describe: List of Chrome command line options to pass on browser start. Can be used to disable CORS to scan protected page: --browserOptions disable-web-security
 ```
+
+-   keepUrlFragment: --keepUrlFragment
+
+```sh
+type: boolean
+describe: To keep the hash fragment in the URLs. If set to false, it will remove the hash fragment from URL. For example, http://www.example.com/#foo will be considered as http://www.example.com. Default value is false.
```
DaveTryon, line 298: This only applies when crawling. If http://www.example.com/#foo is passed in, that page (with the routing) will be scanned. That's the current behavior so it's a breaking change if we alter it. We might want to try to capture that in the documentation, but I'm not sure how we'd want to express it

DaveTryon (approved):
Validated SPA crawling locally. Works well

packages/cli/README.md
v-viyada: In the Readme documentation, new parameter is mentioned for Crawl and batch scan sections and not for single Url scan. The new parameter will not make any difference if it is passed or not for single Url scan.

packages/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "accessibility-insights-scan",
-    "version": "2.4.6",
+    "version": "2.4.7",
```
DaveTryon, line 3: Normally, adding new functionality (which this does) would merit a bump to the minor version instead of the patch version
```suggestion
    "version": "2.5.0",
```

v-viyada: updated

PR #2525, opened 2024-02-15
fix null response for navigation of URLs with hash fragment

v-viyada:
#### Details

crawlee page.goto function used for navigation will always give null response for URLs with hash fragment as it treats hash navigation as single page reload.

page.goto returns response when running locally. If we build and create npm package for cli and then do npm install for package it gives null response while navigating for URLs with hash fragment. This PR is to handle the null check for hash URLs and do not mark it as navigation error.

Steps to Test
1. yarn build
2. yarn run pack
3. npm install cli.tgz (generated by above step)
4. test the crawling command.

##### Motivation

[Feature 2114928](https://dev.azure.com/mseng/1ES/_workitems/edit/2114928): ADO Extension should be able to crawl hash-routing SPA apps

#### Pull request checklist

### DIFF
--- a/packages/crawler/src/page-handler/page-navigation-hooks.ts
+++ b/packages/crawler/src/page-handler/page-navigation-hooks.ts
@@ -40,7 +40,8 @@ export class PageNavigationHooks {
             });
         }
 
-        // Validate HTTP response, hashUrls will have null response from page.goto navigation as crawler treats it same page load
+        // Validate HTTP response
+        // hashUrls will have null response as crawlee's page.goto treats it as same page load
         const responseError = isHashUrl && response === null ? undefined : this.pageResponseProcessor.getResponseError(response);
         if (responseError !== undefined) {
             await onNavigationError(responseError);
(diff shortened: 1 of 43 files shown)


packages/crawler/src/page-handler/page-navigation-hooks.ts
```diff
@@ -30,16 +30,18 @@ export class PageNavigationHooks {
...
             await onNavigationError({
                 errorType: 'NavigationError',
                 message: 'Unable to get a page response from the browser.',
                 stack: new Error().stack,
             });
         }
 
-        // Validate HTTP response
-        const responseError = this.pageResponseProcessor.getResponseError(response);
+        // Validate HTTP response, hashUrls will have null response as crawler treats it same page load
```
DaveTryon, line 43: Suggestion, since we're in the crawler package, it's probably appropriate to be specific about the source of this behavior
```suggestion
        // Validate HTTP response
        // hashUrls will have null response as crawlee's page.goto treats it as same page load
```

DaveTryon (approved):
Looks good! Made one suggestion about a comment. No need to re-review.

packages/crawler/src/page-handler/page-navigation-hooks.ts
v-viyada: Done

PR #2550, opened 2024-05-31
Revert "fix: updated install script to fix test website and apim serv…

v-viyada:
functional test cases started failing in the deployment after https://github.com/microsoft/accessibility-insights-service/pull/2545 so reverting it. 

This reverts commit 12713262f19003bee37ff4db6ed7ac9a83bd3bed.

#### Pull request checklist


alisonlomaka:
Is there any adverse effect from reverting the prior change? Do we need to revisit that change and find another way to implement the same behavior?

v-viyada:
> Is there any adverse effect from reverting the prior change? Do we need to revisit that change and find another way to implement the same behavior?

Yes, we need to revisit that change and find a way to fix deployment for a completely new resource group.
I reverted this change to unblock the deployments to existing resource group.

alisonlomaka (approved):

